PR #8, opened 2022-02-24
chore: fixed projen deps

voho:
removes aws-cdk, etc


package.json
```diff
@@ -52,21 +50,19 @@
...
     "typescript": "^4.5.5"
   },
   "peerDependencies": {
-    "@aws-cdk/core": "^1.123.0",
     "constructs": "^3.2.27",
     "monocdk": "^1.123.0"
   },
   "dependencies": {
-    "@aws-cdk/core": "^1.123.0",
     "monocdk": "^1.123.0"
```
echeung-amzn, line 66: Need to remove this too, it should only be in peer/dev deps.

echeung-amzn (changes requested):

package.json
voho: I do not see it locally. Maybe i updated the PR in a wrong way. Let me check.

voho: Is it still there? https://github.com/cdklabs/cdk-monitoring-constructs/blob/4b9118b4104698803c300b6b20925ec07757a59c/package.json

voho: Removed it as well.

echeung-amzn (approved):

Issue #18, opened 2022-02-25
Packaging issues


echeung-amzn:
Tracking some issues with the npm tarball:

- [x] TS files are included
- [x] Lambda assets aren't transpiled

voho:
I am not 100% sure, but I think it could be these config differences in tsconfig:

Desired setting:

```
  "include": [
    "**/*.ts"
  ],
  "exclude": [
    "node_modules",
    "dist/**/*.ts"
  ],
```

Current setting:

```
  "include": [
    "lib/**/*.ts"
  ],
  "exclude": [
    "node_modules"
  ],
```

echeung-amzn:
Addressed by #23 and #24.

PR #24, opened 2022-02-25
chore: convert Lambda handlers to vanilla JS

echeung-amzn:
Partially addresses #18

We don't necessarily need JSII (and therefore TS) to process these.


voho (approved):
Approving for now to unblock. Would it help to move these into `lib`?

echeung-amzn:
> Would it help to move these into `lib`?

That'd likely make it more complicated because we don't actually want to deal with `aws-sdk` dependencies.

PR #34, opened 2022-03-02
feat: Add scope monitoring with aspects

ayush987goyal:
Ref: https://github.com/cdklabs/cdk-watchful/blob/main/src/aspect.ts

The concept of `MyMonitoringOptions` for every monitoring is needed because `Omit` is not supported by JSII. Please refer this issue and the comment on the linked PR for the way to handle that.
https://github.com/aws/jsii/issues/1707

Opened an issue to add aspects for other left-out resources: https://github.com/cdklabs/cdk-monitoring-constructs/issues/41


echeung-amzn (commented):
This seems technically okay, but it also seems like it'd be unwieldy to update as more things are added.

lib/facade/MonitoringAspect.ts
```diff
@@ -0,0 +1,55 @@
...
+import * as apigwv2 from "monocdk/aws-apigatewayv2";
+import * as acm from "monocdk/aws-certificatemanager";
+
+import { MonitoringFacade } from "./MonitoringFacade";
+
+export interface MonitoringAspectProps {
+  /**
+   * Automatically monitor ACM Certificats in the scope
+   */
+  readonly acm?: boolean;
```
voho, line 12: default is missing

lib/facade/MonitoringAspect.ts
```diff
@@ -0,0 +1,55 @@
+import { IConstruct, IAspect } from "monocdk";
+import * as apigw from "monocdk/aws-apigateway";
+import * as apigwv2 from "monocdk/aws-apigatewayv2";
+import * as acm from "monocdk/aws-certificatemanager";
+
+import { MonitoringFacade } from "./MonitoringFacade";
+
+export interface MonitoringAspectProps {
+  /**
+   * Automatically monitor ACM Certificats in the scope
```
voho, line 10:
```suggestion
   * Automatically monitor ACM Certificats in the scope
   * @default true
```

lib/facade/MonitoringAspect.ts
```diff
@@ -0,0 +1,55 @@
...
+  ) {}
+
+  public visit(node: IConstruct): void {
+    const monitorAcmCertificate = this.props.acm ?? true;
+    if (monitorAcmCertificate && node instanceof acm.Certificate) {
+      this.monitoringFacade.monitorCertificate({
+        certificate: node,
+        alarmFriendlyName: node.node.path,
+      });
+    }
```
voho, line 43: Long-term, we should probably split each to a separate method? What do you think? Or even create some "node type to monitoring" mapping to make it easier?

lib/facade/MonitoringAspect.ts
```diff
@@ -0,0 +1,55 @@
+import { IConstruct, IAspect } from "monocdk";
+import * as apigw from "monocdk/aws-apigateway";
+import * as apigwv2 from "monocdk/aws-apigatewayv2";
+import * as acm from "monocdk/aws-certificatemanager";
+
+import { MonitoringFacade } from "./MonitoringFacade";
+
+export interface MonitoringAspectProps {
```
voho, line 8: Later we can also add `defaultAcmMonitoringProps?: AcmMonitoringProps` that would be applied to all such constructs. That way, we could add automatic alarms, and so on.

test/facade/MonitoringAspect.test.ts
```diff
@@ -0,0 +1,49 @@
...
+import * as apigwv2 from "monocdk/aws-apigatewayv2";
+import * as acm from "monocdk/aws-certificatemanager";
+
+import {
+  DefaultDashboardFactory,
+  DashboardRenderingPreference,
+  MonitoringFacade,
+} from "../../lib";
+
+test("snapshot test: MonitoringAspect", () => {
```
voho, line 13: Even this might grow very long. What about creating one stack in a helper method, and then having one test case for each construct type?

lib/facade/MonitoringAspect.ts
voho: Thinking even bigger:

What about extending each Monitoring with Aspect, so that each monitoring would be able to identify its own construct? And the we would have the MonitorAllAspect, which would just forward the node to all known aspects?

ayush987goyal: Added the 1st suggestion in the latest change.
As per the 2nd one, we would need to see if that is even possible since each monitoring does not have access to the instantiated monitoring facade.

voho (approved):

lib/facade/MonitoringFacade.ts
```diff
@@ -615,4 +616,14 @@ export class MonitoringFacade extends MonitoringScope {
     this.addSegment(segment, props);
     return this;
   }
+
+  /**
+   * Monitor all the resources in the given scope.
```
echeung-amzn, line 621: Should we mark this as `@experimental`, or does the current approach feel finalized?

lib/facade/aspect-types.ts
```diff
@@ -0,0 +1,43 @@
...
+   * Automatically monitor ACM Certificates in the scope
+   * @default true
+   */
+  readonly acm?: boolean;
+
+  /**
+   * Default properties for ACM Certificates Monitoring
+   * @default - none
+   */
+  readonly defaultAcmMonitoringProps?: CertificateManagerMonitoringOptions;
```
echeung-amzn, line 18: Would this be more manageable if we introduce a generic interface like

```ts
interface MonitoringAspectType<T> {
  enabled: boolean;
  props?: T;
}
```

And then use it like 
```ts
{
  readonly acm?: MonitoringAspectType<CertificateManagerMonitoringOptions>;
  readonly apiGateway?: MonitoringAspectType<ApiGatewayMonitoringOptions>;
  // ...
}
```

lib/facade/MonitoringFacade.ts
ayush987goyal: Yes I think we can mark it experimental till we feel confident and/or get feedback.

lib/facade/aspect-types.ts
ayush987goyal: Yes that look good! Thanks for the suggestion. Will try and see how it pans out.

PR #42, opened 2022-03-05
chore: Update reamdme with scope monitoring

ayush987goyal:
chore: Update readme with scope monitoring

Also making the `enabled` prop optional.

follow up to #34

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,28 @@ This is a general procedure on how to do it:
 
 Both of these monitoring base classes are dashboard segments, so you can add them to your monitoring by calling `.addSegment()`.
 
+### Monitoring Scopes
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -209,6 +209,28 @@ This is a general procedure on how to do it:
...
+```ts
+monitoring.monitorScope(stack);
+```
+
+You can also specify default alarms for any specific resource and disable automatic monitoring for it as well.
+
+```ts
+monitoring.monitorScope(stack, {
+  lambda: {
+    props: {
```
voho, line 225: Nit: maybe we could call them defaultProps?

voho (approved):

README.md
echeung-amzn: That implies that they'd be configurable individually too, which doesn't seem to be the case with the current API.

PR #59, opened 2022-03-08
chore: only run upgrade deps workflow once weekly [skip ci]

echeung-amzn:
This is up for debate, but daily dependency updates (and therefore new releases) seem like overkill. 

This also increases the risk/likelihood of bringing in new malicious dependency versions (e.g., what happened with `colors`).


ayush987goyal (approved):
I think this should be alright since if we had to upgrade some version on priority, we can always manually run the workflow (thanks to `workflow_dispatch`) whenever we wany.

PR #74, opened 2022-03-14
chore: use action to get release by current SHA for Slack notification [skip ci]

echeung-amzn:
The previous attempted [failed](https://github.com/cdklabs/cdk-monitoring-constructs/runs/5530778333?check_suite_focus=true) since the action tries to get a release for a given tag, except the workflow is actually triggered on push, not by a tag.

This switches to [cardinalby/git-get-release-action](https://github.com/cardinalby/git-get-release-action), which defaults to getting a release based on the current commit SHA.

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -210,7 +210,9 @@ jobs:
     steps:
       - name: Get release
         id: get_release
-        uses: bruceadams/get-release@v1.2.3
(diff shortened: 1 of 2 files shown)


ayush987goyal (commented):
Did you run `yarn projen` post this change? Shouldn't it also update the workflow file?

Issue #48, opened 2022-03-05
[dynamo] addSystemErrorCountAlarm broken on CDK 1.135.0+


voho:
[This](https://github.com/aws/aws-cdk/pull/17738) change in 1.135.0 added some missing operations in the DynamoDB enum, which makes `metricSystemErrorsCount` exceed the limit of 10 alarms in metric math.

echeung-amzn:
Would it potentially make sense to create the metrics in chunks of 10, then have an alarm based on a composite alarm of alarms for those chunked metrics?

PR #83, opened 2022-03-24
feat!: upgrade to CDK v2

### LINKED ISSUE #37: CDK v2 support
We should migrate to `aws-cdk-lib` from `monocdk` when appropriate. This would likely result in a v1 release of this library.

echeung-amzn:
Closes #37

~~This is actually blocked by #48, so we should address that separately first.~~

Of note:

- CDK v2 doesn't bundle experimental APIs in, so they're separate `@aws-cdk` modules. This makes the peer dependencies quite awkward.
- `dimensions` is no longer accessible in the `Metric` props, forcing us to use `dimensionsMap`, which is typed to only allow `string` values. ~~I'm not entirely sure if forcibly putting non-string values works properly.~~
- The `@aws-cdk/aws-secretsmanager:parseOwnedSecretName` feature flag is removed in CDK v2.
- It's otherwise mostly just import changes

Relevant docs:

- https://docs.aws.amazon.com/cdk/v2/guide/migrating-v2.html

---


lib/monitoring/aws-dynamo/DynamoTableMetricFactory.ts
```diff
@@ -98,7 +98,7 @@ export class DynamoTableMetricFactory {
       'MetricName="SuccessfulRequestLatency"',
       {
         TableName: this.table.tableName,
-        Operation: undefined,
+        Operation: undefined as unknown as string,
```
ayush987goyal, line 101: We would need to see if deployment passes with this. Could you try to deploy this change from a test package to see if everything works fine?

ayush987goyal (approved):
Overall the change looks good to me. Just one comment about testing once by deploying the change (similar to the internal test package).

Thanks a lot for working on this big change!!

lib/monitoring/aws-dynamo/DynamoTableMetricFactory.ts
echeung-amzn: I deployed it to a test account and it seems to work fine. This is the metric that ends up in the dashboard:

```
    "metrics": [
        [ { "label": " ", "expression": "SEARCH('{AWS/DynamoDB,TableName,Operation} TableName=\"SampleTable\" MetricName=\"SuccessfulRequestLatency\"', 'Average', 300)", "region": "us-east-1" } ]
    ],
```

Which matches up with what's generated with the older monocdk-based codebase.

ayush987goyal (approved):

voho (approved):
LGTM! However, there might still be some changes that are not covered, I think it might be the case of WAF.

echeung-amzn:
> there might still be some changes that are not covered, I think it might be the case of WAF.

You mean in terms of pre-existing code coverage?

voho:
> 

I mean that these would need a migration to CDK v2, since they include import changes, right?

echeung-amzn:
This has been rebased on top of the current `HEAD`, so I'll go ahead and allow this to be merged.

Issue #76, opened 2022-03-15
[waf] Support for AWS WAF


rahul0705:
### Feature scope

AWS WAF

### Describe your suggested feature

AWS Offers Several different firewall tools. Having Web Application Firewall being monitored in this library would be greatly helpful.

echeung-amzn:
https://docs.aws.amazon.com/waf/latest/developerguide/monitoring-cloudwatch.html

Can you describe which metrics would be useful for you to monitor and alarm on?

voho:
Just curious - what does this metricName parameter do? I see the metric names are specified already in the doc. What changes when you modify it?

rahul0705:
Based on what I've seen `metricName` allows you to specify custom names for the rule or WAF itself.

As far as API to define monitoring, to keep consistent with other APIs it probably would be a good idea to either take the ACL directly (if we can query the `metricNames` assigned for that ACL) its also worth noting that ACLs can be assigned to an AWS resource, so maybe its possible to "automagically" determine the ACL metrics for services that have them tied already if the API supports it (i.e. API Gateway may have an ACL so `.monitorAPIGateway()` can automatically pull ACL metrics in?, or similar with CloudFront?)

voho:
We can take it step by step, e.g. start with something simple and keep extending it.

I have been researching some existing solutions and it seems that people are usually just monitoring allowed vs. blocked requests, typically for ALL rules in the given ACL.

So, what is the basic unit we are monitoring here... is it ACL + subset of rules? Or just list of rules and ACL separately?

rahul0705:
Probably the latter as it seems from what I've seen that most ACLs will contain very very similar rules, we may also want to include count as well since that will tell the user if a request matched a rule but was not blocked (generally used for testing new rules?)

echeung-amzn:
Basic support was added in #89. It'd be best to create a new issue detailing enhancements in more depth if desired.

PR #89, opened 2022-03-26
feat(waf): add basic WAF monitoring

voho:
Related #76 

---


voho:
We can extend this later, this is just a start.

ayush987goyal (commented):
The changes looks good to me!

Do you think it is worth adding aspect monitoring for it similar to what we do for glue jobs [here](https://github.com/cdklabs/cdk-monitoring-constructs/blob/main/lib/facade/MonitoringAspect.ts#L206-L214) via the [CfnWebACL](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-waf.CfnWebACL.html) L1 construct?

voho:
> The changes looks good to me!
> 
> Do you think it is worth adding aspect monitoring for it similar to what we do for glue jobs [here](https://github.com/cdklabs/cdk-monitoring-constructs/blob/main/lib/facade/MonitoringAspect.ts#L206-L214) via the [CfnWebACL](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-waf.CfnWebACL.html) L1 construct?

I was thinking about it... I think in our case, the CfnWebACL probably makes sense, but to be honest, I am not very familiar with the WAF data model. I can add it, although I expect we might have different additions for WAF, e.g. rule-specific monitoring, and so on.

echeung-amzn (approved):

echeung-amzn:
Tentatively added `do-not-merge` label in case @ayush987goyal had any pending thoughts.

ayush987goyal (approved):

Issue #25, opened 2022-02-25
Deploy an integration test stack to CloudFormation on push


echeung-amzn:
It'd be nice to automatically deploy to an AWS account to ensure things are working as intended.

ayush987goyal:
Some thoughts on this:

1. We could run a workflow on each PR to deploy the current change to CFN.
2. This could be done via creating a CDK app in the tests directory and executing it via an npm script.
3. The dashboard created via this CFN stack can be made public as described [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch-dashboard-sharing.html#share-cloudwatch-dashboard-public). Need to figure out how to enable public sharing via CFN which doesn't seem to be possible.
4. The problem with the above is should we deploy to a new stack for each PR v/s the same stack gets deployed. 
  * If it is different per PR, then how do we clean it up? 
  * If it is the same stack, how do we roll it back if the PR gets closed? Merged PR is not a problem because we would deploy again in release workflow

CaymonSullivan:
> The dashboard created via this CFN stack can be made public as described [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch-dashboard-sharing.html#share-cloudwatch-dashboard-public). Need to figure out how to enable public sharing via CFN which doesn't seem to be possible.

I think this could be made into an L3 construct in the CloudWatch CDK. Looking into the requests CW makes when setting this value in the AWS Console, it's creating a Cognito Identity pool with public access then configuring the Dashboard to use that. Need to dive deeper on how the dashboard is updated, or if it's just an account wide setting.

ayush987goyal:
Possibly also add `integ-runner ` integration if and when it gets open sourced:

https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/integ-runner/README.md

PR #122, opened 2022-04-16
feat(ec2): allow users to specify EC2 instance IDs to monitor

### LINKED ISSUE #43: [ec2] add filter of instance IDs
As a customer:

> I would just prefer a property under .monitorEC2instances that would allow me to specify instance ID's or something along those lines. I didn't deploy the instances via an ASG, so it seems like an all or nothing approach using this construct.

voho:
Allows users to specify instance IDs in the EC2 monitoring and combine them with ASG property.

Fixes #43 

---

### DIFF
--- a/lib/monitoring/aws-ec2/EC2MetricFactory.ts
+++ b/lib/monitoring/aws-ec2/EC2MetricFactory.ts
@@ -5,7 +5,7 @@ import { MetricFactory, MetricStatistic } from "../../common";
 
 const EC2Namespace = "AWS/EC2";
 
-export interface IECMetricFactoryStrategy {
+export interface IEC2MetricFactoryStrategy {
   createMetrics(
     metricFactory: MetricFactory,
     metricName: string,
@@ -16,7 +16,7 @@ export interface IECMetricFactoryStrategy {
 /**
  * Creates a single metric for the whole ASG.
  */
-class AutoScalingGroupStrategy implements IECMetricFactoryStrategy {
+class AutoScalingGroupStrategy implements IEC2MetricFactoryStrategy {
   protected autoScalingGroup: IAutoScalingGroup;
 
   constructor(autoScalingGroup: IAutoScalingGroup) {
@@ -44,7 +44,7 @@ class AutoScalingGroupStrategy implements IECMetricFactoryStrategy {
 /**
  * Creates multiple metrics (one for each instance) with an optional ASG filter.
  */
-class SelectedInstancesStrategy implements IECMetricFactoryStrategy {
+class SelectedInstancesStrategy implements IEC2MetricFactoryStrategy {
   protected instanceIds: string[];
   protected autoScalingGroup?: IAutoScalingGroup;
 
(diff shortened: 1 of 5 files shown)


lib/monitoring/aws-ec2/EC2MetricFactory.ts
```diff
@@ -1,30 +1,158 @@
...
+import { DimensionsMap, IMetric } from "aws-cdk-lib/aws-cloudwatch";
 
 import { MetricFactory, MetricStatistic } from "../../common";
 
 const EC2Namespace = "AWS/EC2";
 
-export class EC2MetricFactory {
-  protected readonly metricFactory: MetricFactory;
-  protected readonly autoScalingGroup?: IAutoScalingGroup;
+export interface IECMetricFactoryStrategy {
```
ayush987goyal, line 8:
```suggestion
export interface IEC2MetricFactoryStrategy {
```

ayush987goyal (commented):
Overall it looks good to me! Thanks for making these changes!

1. Do you think it is worthwhile to try to test this by deploying to an actual stack in an account?
2. Do you think we need to mark this change as breaking?

voho:
> Overall it looks good to me! Thanks for making these changes!
> 

Thanks!

> 1. Do you think it is worthwhile to try to test this by deploying to an actual stack in an account?

Agree. I have tried to first extend the snapshot tests to whatever there was before, then I made changes and made sure the tests do not change any snapshot. Finally, I added new test cases with extended behaviour. But still worthwhile to test on a live stack. Any tips on how to do this? With the internal package, this was easy, maybe I can create a burner account and use this branch of the library? Any other ideas?

> 2. Do you think we need to mark this change as breaking?

You mean because of the metric name changes? It would be correct to do so, however, it depends on how many people are relying on the metric names in their overrides. Keeping it open for now, but no problem of doing it.

ayush987goyal:
> Agree. I have tried to first extend the snapshot tests to whatever there was before, then I made changes and made sure the tests do not change any snapshot. Finally, I added new test cases with extended behaviour. But still worthwhile to test on a live stack. Any tips on how to do this? With the internal package, this was easy, maybe I can create a burner account and use this branch of the library? Any other ideas?

I think we can create a test app (similar to internal tests package) inside an `examples` folder in the repo. We can deploy this stack locally through CLI in personal/burner account and possibly post a screenshot to the PR. Ideally this should be automated by some workflow action as described in this issue: https://github.com/cdklabs/cdk-monitoring-constructs/issues/25.

For this current PR, you can just create the app locally and not push it (for now). We can create that as part of another change. Thoughts?

> You mean because of the metric name changes? It would be correct to do so, however, it depends on how many people are relying on the metric names in their overrides. Keeping it open for now, but no problem of doing it.

Was also thinking on the metric factory actually. But it should be fine I think. @echeung-amzn thoughts on this?

ayush987goyal:
Eventually we would also do something like what `aws-cdk` repo is doing with `integ-runner`. This would be possible if and when the CDK team decides to open source `integ-runner`.

ayush987goyal (approved):

echeung-amzn:
Regarding testing, I did create a [basic sample app](https://github.com/echeung-amzn/sample-cdk-monitoring-app) that I had used to deploy to an internal test account directly, although it's a bit finnicky since `npm-link` seemingly doesn't handle peer dependencies well so I ended up copying/pasting things in `node_modules` directly.

echeung-amzn (approved):

echeung-amzn:
Regarding the breaking change argument, I was actually thinking of whether we should explicitly define what we consider to be part of our "public API" that the semantic versioning revolves around. It'd seem appropriate to ensure the versioning reflects what's available as the top-level facade API, but anything more "internal", despite being exported and extensible as an escape hatch, may have looser guarantees.

PR #163, opened 2022-06-14
feat(cloudfront): allow disabling dashboard widgets based on additional metrics

echeung-amzn:
As pointed out by a user, these metrics are not available by default, so it won't always be relevant.

This is kept as enabled by default to preserve the existing behaviour.

---


API.md
```diff
@@ -8140,6 +8141,22 @@ public readonly distribution: IDistribution;
 
 ---
 
+##### `additionalMetricsEnabled`<sup>Optional</sup> <a name="additionalMetricsEnabled" id="cdk-monitoring-constructs.CloudFrontDistributionMetricFactoryProps.property.additionalMetricsEnabled"></a>
+
+```typescript
+public readonly additionalMetricsEnabled: boolean;
```
voho, line 8147: Should we make this naming consistent with the existing "advanced metrics" in API Gateway?

echeung-amzn: The [docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/viewing-cloudfront-metrics.html#monitoring-console.distributions-additional) refer to them as "additional metrics", so I stuck with that.

ayush987goyal (approved):

PR #164, opened 2022-06-14
feat(widget): add strict graph widgets

voho:
Introducing new specialized graph widget classes. They have customized input props which force user to be consistent in how they use the graph. For example, about the number of axes. It will also force user to provide axis description. One tiny improvement is that for single-metri widgets, we will not show the legend, saving some valuabe space.

These new widgets will be used in the monitoring classes instead of plain GraphWidgets to make sure they are all properly defined.

---


lib/dashboard/widget/StrictGraphWidget.ts
```diff
@@ -0,0 +1,95 @@
...
+  LegendPosition,
+  YAxisProps,
+} from "aws-cdk-lib/aws-cloudwatch";
+
+export interface SingleAxisGraphWidgetProps {
+  readonly title?: string;
+  readonly width: number;
+  readonly height: number;
+  readonly metrics: IMetric[];
+  readonly axis: YAxisProps;
```
echeung-amzn, line 14: We may want to make these more explicit, like `leftYAxis`. There's no confusion as to which axis it is then.

lib/dashboard/widget/StrictGraphWidget.ts
```diff
@@ -0,0 +1,95 @@
...
+ * If there is just one metric, it will hide the legend to save space.
+ * The purpose of this custom class is to make the properties more strict.
+ * It will avoid graphs with no axis, and so on.
+ */
+export class SingleAxisGraphWidget extends GraphWidget {
+  constructor(props: SingleAxisGraphWidgetProps) {
+    super(SingleAxisGraphWidget.transformProps(props));
+  }
+
+  private static transformProps(props: SingleAxisGraphWidgetProps) {
```
echeung-amzn, line 29: Nitpick: add a return type. I should probably get around to revisiting our linting setup.

test/dashboard/widget/StrictGraphWidget.test.ts
```diff
@@ -0,0 +1,55 @@
...
+import { DoubleAxisGraphWidget, SingleAxisGraphWidget } from "../../../lib";
+
+const DummyMetric1 = new Metric({ namespace: "Dummy", metricName: "Metric1" });
+const DummyMetric2 = new Metric({ namespace: "Dummy", metricName: "Metric2" });
+const DummyAxis1: YAxisProps = { min: 0, max: 1 };
+const DummyAxis2: YAxisProps = { min: 100, max: 200 };
+const DummyAnnotation1: HorizontalAnnotation = { value: 42 };
+const DummyAnnotation2: HorizontalAnnotation = { value: 66 };
+
+test("single axis - single metric: snapshot test", () => {
```
echeung-amzn, line 15: Needs failure case tests.

echeung-amzn (changes requested):

lib/dashboard/widget/StrictGraphWidget.ts
voho: Option as well. I wanted to remove the notion of side from the props, but it is true that adding the prefix would result in less changes across the lib (no rename would be needed at most places).

lib/dashboard/widget/StrictGraphWidget.ts
voho: Will do. As for the linting, there is even a ticket for it, since I was not sure what linting config is actually the one used - might be a projen configuration issue.

test/dashboard/widget/StrictGraphWidget.test.ts
voho: OK

lib/dashboard/widget/StrictGraphWidget.ts
```diff
@@ -0,0 +1,95 @@
...
+  readonly leftAnnotations?: HorizontalAnnotation[];
+  readonly rightMetrics: IMetric[];
+  readonly rightAxis: YAxisProps;
+  readonly rightAnnotations?: HorizontalAnnotation[];
+}
+
+/**
+ * Line graph widget with both left and right axes.
+ * The purpose of this custom class is to make the properties more strict.
+ * It will avoid graphs with no axis, and so on.
```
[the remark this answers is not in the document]
voho: In this context, it might be a missing dimension, but if there is not much value in these comments, I can remove them too.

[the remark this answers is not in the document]
echeung-amzn: The comment is useful to call out what makes this stricter without having to go and compare them. Let's explicitly call out the dimensions in that case.

voho:
The instability comes from the Region token, working on it.

echeung-amzn (approved):

PR #166, opened 2022-06-17
fix(dashboards): enforce non-empty dashboardNamePrefix

echeung-amzn:
We currently have this logic in the internal extension library, but this should be enforced at the core library level.

---


voho (approved):
LGTM, thanks for addressing this issue!

PR #173, opened 2022-06-21
Use 'adaptMetric' in all places where it was missing (Lambda, DDB, SQS)

### LINKED ISSUE #172: Lambda: not calling .adaptMetric() in some methods in the metric factory
### Version

latest

https://github.com/cdklabs/cdk-monitoring-constructs/blob/v0.0.42/lib/monitoring/aws-lambda/LambdaFunctionMetricFactory.ts#L156-L161

### Expected behavior

We should wrap the metric created in .adaptMetric() call to normalize the metric period.

### Actual behavior

The metric is returned as-is, which can result in different period than expected.

### Other details

_No response_

voho:
Fixes #172 

---


ayush987goyal (approved):
Not sure why there is no change in snapshots but the build succeeded to I am assuming it was already the case that they included this period?

echeung-amzn:
It's probable that the tests don't have differing periods, so the default 5 minutes still matches up.

echeung-amzn (approved):

PR #174, opened 2022-06-21
feat(docdb): add monitoring for Document DB

### LINKED ISSUE #167: Monitoring support for AWS DocumentDB
### Feature scope

DocumentDB

I just installed v1.10.0 and it does appear that there isn't support for monitoring AWS DocumentDB via this library? I am kindly requesting that this be considered and added, if it's not already been planned.

voho:
Fixes #167 

---


echeung-amzn (changes requested):
Add to `MonitoringAspect` as well?

lib/monitoring/index.ts
```diff
@@ -6,6 +6,7 @@ export * from "./aws-billing";
 export * from "./aws-cloudfront";
 export * from "./aws-cloudwatch";
 export * from "./aws-codebuild";
+export * from "./aws-documentdb";
```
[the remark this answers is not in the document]
voho: OK

lib/monitoring/aws-documentdb/DocumentDbMetricFactory.ts
```diff
@@ -0,0 +1,95 @@
...
+  getLatencyTypeStatistic,
+  LatencyType,
+  MetricFactory,
+  MetricStatistic,
+} from "../../common";
+
+const DocumentDbNamespace = "AWS/DocDB";
+
+export interface DocumentDbMetricFactoryProps {
+  readonly clusterIdentifier: string;
```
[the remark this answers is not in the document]
voho: If such construct already exists, then I will go for it! I was not aware.

lib/monitoring/aws-documentdb/DocumentDbMetricFactory.ts
```diff
@@ -0,0 +1,95 @@
...
+    props: DocumentDbMetricFactoryProps
+  ) {
+    this.metricFactory = metricFactory;
+    this.dimensionsMap = {
+      DBClusterIdentifier: props.clusterIdentifier,
+    };
+  }
+
+  metricAverageCpuUsageInPercent() {
+    return this.metric("CPUUtilization", MetricStatistic.AVERAGE, "CPU Usage");
```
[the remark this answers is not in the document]
voho: I have created a burner account with docdb to check the metrics, will verify the datapoints are meaningful - the datapoints should be populated by now.

The `adaptMetric` is done in the private `metric` method, so we should be covered.

lib/monitoring/aws-documentdb/DocumentDbMonitoring.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.usageAnnotations = [];
+
+    const metricFactory = new DocumentDbMetricFactory(
+      scope.createMetricFactory(),
+      props
+    );
+
+    this.cpuUsageMetric = metricFactory.metricAverageCpuUsageInPercent();
+    this.readLatencyMetric = metricFactory.metricAverageCpuUsageInPercent();
+    this.writeLatencyMetric = metricFactory.metricAverageCpuUsageInPercent();
```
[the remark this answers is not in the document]
voho: Oops. Great catch.

voho:
> Add to `MonitoringAspect` as well?

Happy to, since there is this L2 construct. However, I am not sure if there is a way to distinguish different cluster types, e.g. DocDB cluster from RDS cluster. Will check what the exact type of differentiator is.

lib/monitoring/aws-docdb/DocumentDbMetricFactory.ts
```diff
@@ -0,0 +1,121 @@
...
+  MetricStatistic,
+} from "../../common";
+
+const DocumentDbNamespace = "AWS/DocDB";
+
+export interface DocumentDbMetricFactoryProps {
+  /**
+   * database cluster identifier (either this or `cluster` need to be specified)
+   */
+  readonly clusterIdentifier?: string;
```
ayush987goyal, line 18: Do we really need to support both? We typically are not doing that for other monitorings.

lib/monitoring/aws-docdb/DocumentDbMetricFactory.ts
```diff
@@ -0,0 +1,121 @@
...
+  ) {
+    return this.metricFactory.createMetric(
+      metricName,
+      statistic,
+      label,
+      this.dimensionsMap,
+      undefined,
+      DocumentDbNamespace
+    );
+  }
```
ayush987goyal, line 120: Would this need to be wrapped with adapt?

lib/monitoring/aws-docdb/DocumentDbMetricFactory.ts
```diff
@@ -0,0 +1,121 @@
...
+      return props.clusterIdentifier;
+    } else if (
+      props.cluster !== undefined &&
+      props.clusterIdentifier === undefined
+    ) {
+      return props.cluster.clusterIdentifier;
+    } else {
+      throw new Error(
+        "Specify either 'clusterIdentifier' or 'cluster', but not both."
+      );
```
ayush987goyal, line 53: This else block also covers the case where neither of them are specified. The message in that case would be misleading.
We should add two different error messages when both or neither are provided

ayush987goyal (changes requested):

lib/monitoring/aws-docdb/DocumentDbMetricFactory.ts
voho: I know, but I was thinking about two reasons:

1) The cluster L2 constructs are quite new and users might not be using them yet (they might only have cluster ID available)
2) The other construct (RDS) was using identifier and should be using cluster too - here I wanted to be backwards compatible and consistent

Not sure it is the best decision, what do you think?

lib/monitoring/aws-docdb/DocumentDbMetricFactory.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export interface DocumentDbMetricFactoryProps {
+  /**
+   * database cluster identifier (either this or `cluster` need to be specified)
+   */
+  readonly clusterIdentifier?: string;
+  /**
+   * database cluster (either this or `clusterIdentifier` need to be specified)
+   */
+  readonly cluster?: DatabaseCluster;
```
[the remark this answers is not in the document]
voho: Indeed!

lib/monitoring/aws-docdb/DocumentDbMetricFactory.ts
voho: OK - I thought it was clear enough, but we can raise the bar here and be very specific.

lib/monitoring/aws-docdb/DocumentDbMetricFactory.ts
ayush987goyal: 1. Even though it is new, if it comes under our minimum supported `aws-cdk-lib` version we should use it. For the users who might have just the ID available, they can import the Cluster via [fromDatabaseClusterAttributes](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_docdb.DatabaseCluster.html#static-fromwbrdatabasewbrclusterwbrattributesscope-id-attrs)  which takes [DatabaseClusterAttributes](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_docdb.DatabaseClusterAttributes.html) which inturn only needs the ID as required argument.
2. For RDS, we can keep it the way you have right now (supporting both) but we should mark the ID as depreacated and instruct them to use the IDatabaseCluster instead (same things apply here)

voho: OK, no problem, I can do it like this too.

lib/monitoring/aws-docdb/DocumentDbMetricFactory.ts
voho: cool, will do

ayush987goyal (approved):
Thanks a lot for making these changes!

voho:
Hey folks, thanks for review. Did I miss any change requested? Please let me know :).

echeung-amzn (approved):

PR #200, opened 2022-07-21
feat: add minMetricSamplesToAlarm property

### LINKED ISSUE #198: Allows users specify minimal sample size below which the alarm will be suppressed and considered unreliable
### Feature scope

Alarming

The need for this feature is most visible when alarming on the AVERAGE, P90, P99... of some metric, which might contain sparse datapoints and hence decrease the reliability of the aggregated value.

Example:
You have an alarm on the P99 Latency metric. But during nights the TPS is going very low, hence you don't get enough records to reliably understand if there is an issue increasing latency or it's because of regular cache misses, retries, etc... If during 1 minute you got only a few calls to the service, the P99 latency won't be reliable reprepsentation of the health of the system. Hence we don't want to alarm on it.

With the current setup of the monitoring constructs, this is not easily possible. We have to create the composite alarms manually, add the actions, deal with the sample size metrics, etc...

By adding this to the library, we'll allow users to create alarms that are closer to their needs.

Definition of done:
- By adding a field to the alarms definition, the users can define minimal sample size after which they want the alarm to be enabled
- The change should be backwards compatible

voho:
Fixes #198 

---


lib/common/alarm/AlarmFactory.ts
```diff
@@ -496,20 +517,62 @@ export class AlarmFactory {
...
+
+    let alarm: AlarmBase = primaryAlarm;
+
+    // create composite alarm for min metric samples (if defined)
+
+    if (props.minMetricSamplesToAlarm) {
+      const metricSampleCount = metric.with({ statistic: MetricStatistic.N });
+      const noSamplesAlarm = metricSampleCount.createAlarm(
+        this.alarmScope,
+        alarmName + "-NoSamples",
```
echeung-amzn, line 549:
```suggestion
        `${alarmName}-NoSamples`,
```

lib/common/alarm/AlarmFactory.ts
```diff
@@ -496,20 +517,62 @@ export class AlarmFactory {
...
+        {
+          alarmName: alarmName + "-NoSamples",
+          alarmDescription: "The metric does not have enough samples.",
```
echeung-amzn, line 552: Maybe something like "${Metric name} has insufficient samples."?

lib/common/alarm/AlarmFactory.ts
echeung-amzn: Likewise for other string concatenations.

test/common/alarm/AlarmFactory.test.ts
```diff
@@ -193,9 +193,9 @@ test("addAlarm: period override is propagated to alarm metric", () => {
     alarmNameSuffix: "TwoHoursPeriod",
     period: Duration.hours(2),
   });
-  const alarm1hConfig = alarm1h.alarm.metric.toMetricConfig();
+  const alarm1hConfig = (alarm1h.alarm as any).metric.toMetricConfig();
```
echeung-amzn, line 196: Could cast as an `Alarm` for better type checking?

echeung-amzn (changes requested):

lib/common/alarm/AlarmFactory.ts
voho: Will fix

lib/common/alarm/AlarmFactory.ts
voho: Yeah, maybe better to put more details there.

test/common/alarm/AlarmFactory.test.ts
voho: Good catch, will change.

echeung-amzn (approved):

PR #206, opened 2022-08-02
feat(lambda): alarming and dashboard support for success percentage

zorbasd:
- Add lambda success percentage metric.
- Add lambda success percentage alarm.
- Add lambda success percentage dashboard widget.

Fixes #

---


echeung-amzn (changes requested):
LGTM, but can you also mention this in the features table in `README.md`?

voho:
Thanks for the contribution! How does the lambda monitoring section look like after the change? Do you have any screenshot? If we are modifying the defaults, we should be super careful not to break the layout for the existing users.

niebloomj:
Would love to see some screenshots here

echeung-amzn:
Closing abandoned PR.

PR #219, opened 2022-08-18
feat(secretsmanager): provide method to interact with metrics publisher Lambda function

echeung-amzn:
Based on a customer request since they want to monitor the function too.

---


lib/monitoring/aws-secretsmanager/SecretsManagerMetricsPublisher.ts
```diff
@@ -20,7 +20,7 @@ import { SecretsManagerSecretMetricFactory } from "./SecretsManagerSecretMetricF
 
 export class SecretsManagerMetricsPublisher extends Construct {
   private static instances: Record<string, SecretsManagerMetricsPublisher> = {};
-  private readonly lambda: IFunction;
+  public readonly lambda: IFunction;
```
voho, line 23: LGTM! Just a nit - we can remove `public` and keep it default, right?

voho (approved):

PR #230, opened 2022-08-24
feat: make the minMetricSamplesToAlarm available for any alarm

### LINKED ISSUE #229: Feature suggestion: Add `minMetricSamplesToAlarm` to `ErrorRateThreshold`
I would like to suggest adding `minMetricSamplesToAlarm` to `ErrorRateThreshold` (or to its parent class `CustomAlarmThreshold`), so that one can add `minMetricSamplesToAlarm` to the `addFaultRateAlarm` entry in the props for `MonitoringFacade.monitorLambdaFunction()`.

There is no way I can find to specify either the requirement for statistical significance (`evaluateLowSampleCountPercentile: false` results in a CFN deployment failure saying that it is only applicable to percentile based metrics (which I had originally guessed that a rate would be, but it appears not) or a minimum number of samples when using
```
monitoringFacade.monitorLambdaFunction({
    ...
    addFaultRateAlarm: {
    },
    ...
});
```

voho:
Fixes #229 

---

### DIFF
--- a/lib/common/alarm/CustomAlarmThreshold.ts
+++ b/lib/common/alarm/CustomAlarmThreshold.ts
@@ -40,6 +40,18 @@ export interface CustomAlarmThreshold {
   readonly alarmDescriptionOverride?: string;
 
   /**
+   * Specifies how many samples (N) of the metric is needed to trigger the alarm.
+   * If this property is specified, an artificial composite alarm is created of the following:
+   * <ul>
+   * <li>The original alarm, created without this property being used; this alarm will have no actions set.</li>
+   * <li>A secondary alarm, which will monitor the same metric with the N (SampleCount) statistic, checking the sample count.</li>
+   * </ul>
+   * The newly created composite alarm will be returned as a result, and it will take the original alarm actions.
+   * @default - default behaviour - no condition on sample count will be added to the alarm
+   */
+  readonly minMetricSamplesToAlarm?: number;
+
+  /**
    * This allows user to attach custom values to this alarm, which can later be accessed from the "useCreatedAlarms" method.
    *
    * @default - no tags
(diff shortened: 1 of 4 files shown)


lib/common/alarm/CustomAlarmThreshold.ts
```diff
@@ -39,6 +39,18 @@ export interface CustomAlarmThreshold {
    */
   readonly alarmDescriptionOverride?: string;
 
+  /**
+   * Specifies how many samples (N) of the metric is needed to trigger the alarm.
+   * If this property is specified, an artificial composite alarm is created of the following:
```
echeung-amzn, line 44:
```suggestion
   * If this property is specified, an composite alarm is created of the following:
```

lib/common/alarm/CustomAlarmThreshold.ts
```diff
@@ -39,6 +39,18 @@ export interface CustomAlarmThreshold {
...
+   * <ul>
+   * <li>The original alarm, created without this property being used; this alarm will have no actions set.</li>
+   * <li>A secondary alarm, which will monitor the same metric with the N (SampleCount) statistic, checking the sample count.</li>
+   * </ul>
+   * The newly created composite alarm will be returned as a result, and it will take the original alarm actions.
```
echeung-amzn, line 49:
```suggestion
   * This composite alarm will be returned as a result and uses the specified alarm actions.
```

echeung-amzn (commented):
Minor wording suggestions.

lib/common/alarm/CustomAlarmThreshold.ts
mavjop: If you remove "artificial", make sure to change the preceding "an" to an "a". :)

echeung-amzn (approved):

Issue #218, opened 2022-08-16
Moving towards composition rather than inheritance for greater modularity


Dicee:
### Feature scope

Everything

### Description

I'm proposing to expose all immutable properties, as well as methods creating new objects (like widgets) without side-effect on the construct, as public rather than protected (for example, exposing `LambdaFunctionMonitoring#faultRateMetric`).

### Use Case

One can extend subclasses of `Monitoring` (e.g. `LambdaFunctionMonitoring`) in order to override the widgets, kind of alarms created etc when the construct doesn't do what the user wants by default. This pattern works fairly well for building a construct that can be reused for several use cases. However, it doesn’t work as well for highly customized monitoring.
 
For example, the lambda monitoring construct really doesn’t cut it for me. Its dashboard is much too detailed for my simple Lambda and most metrics don’t make sense for me, plus I want to add some of my own, and they only apply to a single Lamba in my application, not any other. Just to show how inappropriate the default behaviour is to my use-case:
- **TPS widget:** my lambda gets consistently called at 1 call per minute via a CloudWatch event, so I don't care about TPS
- **Latency widget:** the lambda is asynchronous and will typically execute in seconds, I don't care about latency
- **Error rate widget:** useful, but not as a rate. Since the lambda executes so rarely, I just want to know if each run was successful or not, I don't care about the rate over a period of time. It has to succeed at every run to be healthy.
- **Throttles and provisioned concurrency:** don't care, I don't have any provisioning and it can't throttle as it's called once a minute (unless my latency exceeds a minute with reserved concurrency of 1 but it's impossible because my timeout is 30s)
- **Invocations:** don't care, I know it gets called once a minute
- **Iterator:** not relevant to the way the lambda is triggered
- **Errors:** finally a relevant widget!
- **Business metrics:** I need to add them myself

This is where inheritance and template classes typically fail: they’re too limited, because they impose a template. The more abstract the template, the more customizations are possible, but also the more the abstract class becomes convoluted and hard to make sense of, particularly after several levels of inheritance.
 
To fix these shortcomings, I propose moving towards composition rather than inheritance. Imagine code like this:
 
```typescript
public monitorSnapshotGenerationFunction(fn: SecureFunction) {
    const period = SNAPSHOT_POLLING_RATE;
    const lambdaMonitoring = new ContraCogsLambdaMonitoring(this.monitoring, this.serviceHealthAlarmCollector, {
        lambdaFunction: fn,
[...]

PR #240, opened 2022-09-08
feat: Expose protected attributes and methods (only create*Widget) of all monitoring classes to allow users to build custom monitoring more flexibly via composition rather than inheritance.

Dicee:
feat: Expose protected attributes and methods (only create*Widget) of all monitoring classes to allow users to build custom monitoring more flexibly via composition rather than inheritance.

See https://github.com/cdklabs/cdk-monitoring-constructs/issues/218

---

### DIFF
--- a/API.md
+++ b/API.md
@@ -37474,6 +37474,11 @@ new ApiGatewayMonitoring(scope: MonitoringScope, props: ApiGatewayMonitoringProp
 | <code><a href="#cdk-monitoring-constructs.ApiGatewayMonitoring.createWidgetFactory">createWidgetFactory</a></code> | Creates a new widget factory. |
 | <code><a href="#cdk-monitoring-constructs.ApiGatewayMonitoring.summaryWidgets">summaryWidgets</a></code> | Returns widgets to be placed on the summary dashboard. |
 | <code><a href="#cdk-monitoring-constructs.ApiGatewayMonitoring.widgets">widgets</a></code> | Returns widgets to be placed on the main dashboard. |
+| <code><a href="#cdk-monitoring-constructs.ApiGatewayMonitoring.createErrorCountWidget">createErrorCountWidget</a></code> | *No description.* |
+| <code><a href="#cdk-monitoring-constructs.ApiGatewayMonitoring.createErrorRateWidget">createErrorRateWidget</a></code> | *No description.* |
+| <code><a href="#cdk-monitoring-constructs.ApiGatewayMonitoring.createLatencyWidget">createLatencyWidget</a></code> | *No description.* |
+| <code><a href="#cdk-monitoring-constructs.ApiGatewayMonitoring.createTitleWidget">createTitleWidget</a></code> | *No description.* |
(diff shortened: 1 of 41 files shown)


echeung-amzn (dismissed):
Requires a rebase. It'd be nice to have a example unit test similar to what's given in #218 too, but that can be in a separate PR.

niebloomj:
Do you need help resolving merge conflicts?

Dicee:
Thanks @echeung-amzn ! I'll look into adding a unit test. @niebloomj thanks for offering, if you have permissions feel free to help, I see there are a little more than 30 conflicts in just one file. What I've done shouldn't really impact this file so I think I'll just take the version from main for every conflict.

niebloomj:
> Thanks @echeung-amzn ! I'll look into adding a unit test. @niebloomj thanks for offering, if you have permissions feel free to help, I see there are a little more than 30 conflicts in just one file. What I've done shouldn't really impact this file so I think I'll just take the version from main for every conflict.

I probably don't have permissions lol I just really want this to get pushed through so want to offer whatever I can do to help.

echeung-amzn (approved):
LGTM, but will wait for a second approval before merging since it's a relatively impactful change.

Dicee:
Thanks for the support @echeung-amzn !

ayush987goyal (approved):
LGTM. One callout though, this will expose a lot of internal API which would mean it would be difficult to make opinionated changes in the future. We need to be careful on such changes to ensure that we don't break customers. With great power comes great responsibility!

Dicee:
Makes sense, though as already argued in the corresponding issue, this was already exposed via inheritance, and already leveraged by clients! Thanks all for the reviews and support.

niebloomj (approved):

niebloomj:
I'd like to see a migration document for how we want to handle breaking changes and deprecation warnings given this exposure, even though the exposure was technically already there.

Dicee:
Since they're all very small pieces of logic that are completely independent from one another, I expect it will be quite a simple deprecation strategy. I say deprecation because it's the only kind of relevant change, since the existing API already exposes (and quite dramatically) clients to behavior changes. If anything, people using these methods will be pleased to know their dashboard cannot dramatically change from one version to the next, while those fully relying on preset decisions made for them will have to accept these decisions can change under their feet.

So for deprecation, I was saying, I'd say it's just a matter of marking the field or method as deprecated, and eventually removing it. Since they are independent, having some methods lingering for longer than desired won't be too much of an annoyance as it won't prevent modifying the rest of the code as needed. Overall, I don't feel very concerned about this but let me know if I'm over-simplifying :)

voho (approved):
I am still bit on fence with this, since it will basically expose everything... but if the users find this useful and we can leverage it as well, e.g. for building composition-based structures, lets go with it.

Apart from this, I am more worried about missed monitoring classes and conflicts - are you confident enough we covered everything? We will have to make one more sweek to make sure we did not miss any class and do the same for the internal library.

Dicee:
Thanks for the review! I think I went through all relevant classes, yes. On "conflicts", I'm not sure what you have in mind, could you detail your concern?

Dicee:
Can someone with write access please help pushing this?

Dicee:
Thanks @ayush987goyal

Issue #212, opened 2022-08-10
Can only be used with specific aws-cdk-lib versions due to alpha module peer dependencies


hteek:
### Version

v1.18.4

### Steps and/or minimal code example to reproduce

The peer-dependencies are specified in a way that is not working when using any cdk version higher than 2.18.0.

```
    "@aws-cdk/aws-apigatewayv2-alpha": "^2.18.0-alpha.0",
    "@aws-cdk/aws-appsync-alpha": "^2.18.0-alpha.0",
    "@aws-cdk/aws-redshift-alpha": "^2.18.0-alpha.0",
    "@aws-cdk/aws-synthetics-alpha": "^2.18.0-alpha.0",
```

These only match `2.18.0-alpha.0`, `2.18.0-alpha.1`, `2.18.0-alpha.2` and so on but not `2.19.0-alpha.0` or in my case `2.22.0-alpha.0`

### Expected behavior

Any minor version higher than 2.18.0-alpha.0 can be used.

### Actual behavior

Error: Error: Declared dependency on version ^2.18.0-alpha.0 of @aws-cdk/aws-apigatewayv2-alpha, but version 2.22.0-alpha.0 was found

### Other details

I couldn't find a way to make it work with overrides or resolutions and to make it even worse there is no notation that matches `2.18.0-alpha.0` and `2.19.0-alpha.0` other than `2.18.0-alpha.0 || 2.19.0-alpha.0`.

echeung-amzn:
Assuming you're using npm >= 7, you should be able to configure `legacy-peer-deps` ([docs](https://docs.npmjs.com/cli/v7/using-npm/config#legacy-peer-deps)) to allow for newer versions to be installed, at the risk that the alpha modules' APIs may not match up.

One solution would be to occasionally bump our peer dependencies' versions, but that may be annoying for users.

hteek:
There is no problem when installing the new versions. The problem is that the jsii compiler is not able to compile my construct with the newer versions. Sorry if that was not clear.

```
$ jsii
Error: Error: Declared dependency on version ^2.18.0-alpha.0 of @aws-cdk/aws-apigatewayv2-alpha, but version 2.22.0-alpha.0 was found
    at DependencyResolver.discoverDependencyTree (/Users/h.teek/git/decadia/move37_rentnerportal/node_modules/jsii/lib/project-info.js:131:23)
[...]

PR #241, opened 2022-09-08
chore: remove peer dependency on @aws-cdk/aws-appsync-alpha

echeung-amzn:
Related to #212

This is a bit of a POC that we can remove direct peer dependencies on an alpha module at the cost of some manual interface duplication.

At a high level, we partially copy the interface for the relevant construct with whatever we need for the dimensions. Users can still pass in the alpha L2 constructs since they match up with the partial interfaces.

This allows us some flexibility in updating them, albeit not strongly coupled with actual versioning on the AWS-CDK side. 

---

### DIFF
--- a/lib/monitoring/aws-appsync/AppSyncMetricFactory.ts
+++ b/lib/monitoring/aws-appsync/AppSyncMetricFactory.ts
@@ -1,4 +1,3 @@
-import { GraphqlApi } from "@aws-cdk/aws-appsync-alpha";
 import { DimensionsMap } from "aws-cdk-lib/aws-cloudwatch";
 
 import {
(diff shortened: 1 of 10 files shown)


lib/monitoring/aws-appsync/AppSyncMetricFactory.ts
```diff
@@ -9,6 +8,26 @@ import {
 
 const Namespace = "AWS/AppSync";
 
+/**
+ * Partial copy of [GraphqlApi] from "@aws-cdk/aws-appsync-alpha"
+ *
+ * @experimental Subject to change to match the alpha module.
+ */
+export interface GraphqlApi {
```
ayush987goyal, line 16: How does this work with JSII? The interfacing would not be similar to TypeScript where it infers the members.

echeung-amzn: Good point, it probably doesn't work in languages like Java.

PR #279, opened 2022-12-01
fix(UsageAlarmFactory): allowing addition of multiple alarms from the same class

AakarshanSingla:
Fixes # Addition of multiple alarms from the sameclass

UsageAlarmFactory is creating same AlarmSuffix  for every alarm that is being added in a class which leads to creation of duplicate alarms, as they have same name. This fix appends the UsageType to the alarm name thus allowing users to create alarm on avg, p50,p90... all metrics from the same class.

---

### DIFF
--- a/lib/common/monitoring/alarms/UsageAlarmFactory.ts
+++ b/lib/common/monitoring/alarms/UsageAlarmFactory.ts
@@ -56,6 +56,13 @@ export class UsageAlarmFactory {
     });
   }
 
+  /**
+   * @deprecated Did not allow support of adding multiple metrics from the same class.
+   * @param percentMetric 
+   * @param props 
+   * @param disambiguator 
+   * @returns 
+   */
   addMaxCpuUsagePercentAlarm(
     percentMetric: MetricWithAlarmSupport,
     props: UsageThreshold,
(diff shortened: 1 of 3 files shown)


lib/common/monitoring/alarms/UsageAlarmFactory.ts
```diff
@@ -75,6 +82,31 @@ export class UsageAlarmFactory {
...
 
+  /**
+   * Allows addition of multiple metrics from the same class.
+   * @param percentMetric 
+   * @param usageType Will be added to Alarm name
+   * @param props 
+   * @param disambiguator 
+   * @returns 
+   */
+  addMaxCpuUsagePercentAlarms(
```
echeung-amzn, line 93: This naming doesn't really make sense, it's still only adding a single alarm.

What do you think about adding an additional argument for `usageType` to the existing function and retaining existing behaviour if not specified?

lib/common/monitoring/alarms/UsageAlarmFactory.ts
```diff
@@ -75,6 +82,31 @@ export class UsageAlarmFactory {
     });
   }
 
+  /**
+   * Allows addition of multiple metrics from the same class.
+   * @param percentMetric 
+   * @param usageType Will be added to Alarm name
+   * @param props 
+   * @param disambiguator 
+   * @returns
```
echeung-amzn, line 91: These don't really add value as-is. Either leave them out or fill out the descriptions.

echeung-amzn (changes requested):
Please make sure to run `yarn build` locally first and commit all changes.

PR #280, opened 2022-12-02
fix(UsageAlarmFactory) adding support to add multiple alarms from sam…

AakarshanSingla:
…e class

Fixes #
UsageAlarmFactory is creating same AlarmSuffix for every alarm that is being added in a class which leads to creation of duplicate alarms, as they have same name. This fix appends the UsageType to the alarm name thus allowing users to create alarm on avg, p50,p90... all metrics from the same class.
---


echeung-amzn:
Closing in favour of #279

Issue #307, opened 2023-01-22
DynamoDB Latency Alarm Actions Do Not Adhere to Disambiguator Action Defaults


adelassus:
### Version

v2.1.2

### Steps and/or minimal code example to reproduce

 Latency alarms on DynamoDB tables do not use the `disambiguatorAction` default alarm configuration.
1. Create a `MonitoringFacade` with the `Warning` disambiguator in `alarmFactoryDefaults.disambiguatorAction`  set to an `SnsAlarmActionStrategy` and `alarmFactoryDefaults..action` set to an `OpsItemAlarmActionStrategy`. 
```ts
const monitoringFacade = new MonitoringFacade(scope, id, {
  alarmDefaults: {
    disambiguatorAction: {
      Warning: new SnsAlarmActionStrategy(...),
    },
    action: new OpsItemAlarmActionStrategy(...),
  },
});
```
2. Create latency and capacity alarms for a DynamoDB table:
```ts
monitoringFacade.monitorDynamoTable({
  table: table,
  tableName: tableName,
  ...
  addAverageSuccessfulGetItemLatencyAlarm: {
    Warning: { maxLatency: warningLatency },
  },
  addConsumedReadCapacityAlarm: {
    Warning: {
      maxConsumedCapacityUnits: readCapacityThreshold,
    },
  },
});
```

### Expected behavior

The latency and capacity alarms should be created with SNS actions.

### Actual behavior

The capacity alarms are created with SNS alarms, but the latency alarms are created with the OpsItem actions.

### Other details

The root cause is that In `LatencyAlarmFactory`, the [disambiguator passed into the AlarmFactory](https://github.com/cdklabs/cdk-monitoring-constructs/blob/v2.1.2/lib/monitoring/aws-dynamo/DynamoTableMonitoring.ts#L346) includes the `operation`, which causes [`AlarmFactory.determineAction()`](https://github.com/cdklabs/cdk-monitoring-constructs/blob/v2.1.2/lib/common/alarm/AlarmFactory.ts#L719) to not use the appropriate action from `this.globalAlarmDefaults.disambiguatorAction`.

PR #308, opened 2023-01-22
fix: Fix latency alarm actions to adhere to the disambiguator action defaults

adelassus:
Addresses #307.

---


echeung-amzn:
Thanks for the report + PR! 

Looks like the build's failing due to some missing changes. You'll want to run `yarn build` locally and commit everything that's changed including `API.md` and any updated test snapshots, assuming they look correct.

adelassus:
You're welcome!

I added `API.md` and the build now passes.

I didn't update any of the tests or snapshots as it appears that all the tests are [configured to use the no-op action](https://github.com/cdklabs/cdk-monitoring-constructs/blob/main/test/monitoring/TestMonitoringScope.ts#L25). Therefore, I assumed that you didn't want someone else making such a significant change to the tests and would instead want to do this as a separate effort.

echeung-amzn (approved):

PR #311, opened 2023-01-24
feat: Add region and account parameters as optional global defaults

edisongustavo:
This lets me create cross-account and cross-region monitoring. Currently only the current account is supported.

My PR adds these parameters at the *global level* because I didn't want to go through every resource and add these extra parameters. This creates a limitation where an instance of `MonitoringFacade` would refer only to those specific region/accounts.

This is good enough for me though, so please advise if it's not desired.

---

### DIFF
--- a/API.md
+++ b/API.md
@@ -24429,8 +24429,25 @@ const metricFactoryDefaults: MetricFactoryDefaults = { ... }
 
 | **Name** | **Type** | **Description** |
 | --- | --- | --- |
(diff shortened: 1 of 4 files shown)


echeung-amzn (approved):
This seems like a good way to go about doing it. The logical separation of facades to handle different accounts' metrics seems appropriate.

I'll mark this as `do-not-merge` so that we can get a second review too.

ayush987goyal (approved):

svkurowski (approved):
Looks good to me! Thank you for your contribution.

PR #315, opened 2023-01-27
feat(alarm): Make AlarmNamingStrategy customizable

Mitchell-Cook:
Making the AlarmNamingStrategy configurable. My team wants to add the disambiguator to the alarm dedupe string. This requires custom logic for generating the deduplication string, which is why we want to be able to override this class's functionality.

### DIFF
--- a/test/common/alarm/AlarmFactory.test.ts
+++ b/test/common/alarm/AlarmFactory.test.ts
@@ -15,6 +15,7 @@ import {
   AlarmFactory,
   AlarmFactoryDefaults,
   CompositeAlarmOperator,
+  IAlarmNamingStrategy,
   MetricFactoryDefaults,
   multipleActions,
   noopAction,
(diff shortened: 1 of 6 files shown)


test/common/alarm/AlarmFactory.test.ts
```diff
@@ -409,3 +410,32 @@ test("addAlarm: disambigatorAction takes precedence over default action", () =>
...
+  const alarmLabel = "alarmLabel";
+  const alarmDedupe = "alarmDedupe";
+  const stack = new Stack();
+  const customNamingStrategy: IAlarmNamingStrategy = {
+    // @ts-ignore
+    getName: (props: AlarmNamingInput) => alarmName, 
+    // @ts-ignore
+    getWidgetLabel: (props: AlarmNamingInput) => alarmLabel, 
+    // @ts-ignore
+    getDedupeString: (props: AlarmNamingInput) => alarmDedupe,
```
Mitchell-Cook, line 425: I do get some complaints when running locally that it doesn't like the fact that props is never used, but it is necessary when implementing the interface.

This may be an issue, but curious on your thoughts

Mitchell-Cook: I'll fix this in another Pull Request. I have a better solution.

Mitchell-Cook:
Closed in favor of: https://github.com/cdklabs/cdk-monitoring-constructs/pull/316

PR #321, opened 2023-02-06
feat(ec2): Also plot EBS disk metrics when NITRO instance type is specified

edisongustavo:
EC2 provides different metrics for Nitro-based instances. These metrics have the prefix `EBS`.

These instances either have the `Disk*` or `EBSDisk*` metrics, never both. Because of this I thought in using [metric math](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/using-metric-math.html) where if any of the time series is empty, it would be removed, and then it would display a single metric. This gets trickier when monitoring an ASG because it can have both metrics there, so what should be done with the 2 time series? An average of them? If you think this would be a better idea, please advise so I change the implementation.

---


lib/monitoring/aws-ec2/EC2MetricFactory.ts
```diff
@@ -134,15 +145,28 @@ export interface EC2MetricFactoryProps {
    * @default - no instance filter
    */
   readonly instanceIds?: string[];
+
+  /**
+   * Tells wether the instances are nitro-based. If yes, then uses the metrics with the EBS prefix.
```
echeung-amzn, line 150:
```suggestion
   * Whether the instances are Nitro-based. If Nitro-based, metrics use the EBS prefix.
```

lib/monitoring/aws-ec2/EC2MetricFactory.ts
```diff
@@ -134,15 +145,28 @@ export interface EC2MetricFactoryProps {
...
+   *
+   * @default InstanceFamilyTypes.CLASSIC
+   */
+  readonly instanceFamilyType?: InstanceFamilyTypes;
 }
 
 export class EC2MetricFactory {
   protected readonly metricFactory: MetricFactory;
   protected readonly strategy: IEC2MetricFactoryStrategy;
+  protected instanceFamilyType: InstanceFamilyTypes;
```
echeung-amzn, line 163:
```suggestion
  protected readonly instanceFamilyType: InstanceFamilyTypes;
```

test/monitoring/aws-ec2/EC2Monitoring.test.ts
```diff
@@ -2,70 +2,90 @@ import { Stack } from "aws-cdk-lib";
...
-      "DummyASG",
-      "DummyASG"
-    ),
-  });
-
-  addMonitoringDashboardsToStack(stack, monitoring);
-  expect(Template.fromStack(stack)).toMatchSnapshot();
-});
+function enumKeys<O extends object, K extends keyof O = keyof O>(obj: O): K[] {
+  return Object.keys(obj).filter((k) => Number.isNaN(+k)) as K[];
```
echeung-amzn, line 10: The filtering seems redundant.

test/monitoring/aws-ec2/EC2Monitoring.test.ts
```diff
@@ -2,70 +2,90 @@ import { Stack } from "aws-cdk-lib";
...
+}
+
+test.each(enumKeys(InstanceFamilyTypes))(
```
echeung-amzn, line 13: You could just iterate over the enum values once and do the tests rather than iterating repeatedly for each test.

test/monitoring/aws-ec2/EC2Monitoring.test.ts
```diff
@@ -2,70 +2,90 @@ import { Stack } from "aws-cdk-lib";
...
+  "snapshot test: all instances, no alarms. Instance family type: %p",
+  (instanceFamilyType) => {
+    const stack = new Stack();
+
+    const scope = new TestMonitoringScope(stack, "Scope");
+
+    const monitoring = new EC2Monitoring(scope, {
+      alarmFriendlyName: "EC2",
+      instanceFamilyType: InstanceFamilyTypes[instanceFamilyType],
```
echeung-amzn, line 22: It'd be good to capture default behaviour in a separate test too.

lib/monitoring/aws-ec2/EC2MetricFactory.ts
```diff
@@ -204,10 +228,26 @@ export class EC2MetricFactory {
   }
 
   private createMetrics(metricName: string, statistic: MetricStatistic) {
-    return this.strategy.createMetrics(
-      this.metricFactory,
-      metricName,
-      statistic
+    let metrics = new Array<string>();
+    if (metricName.startsWith("Disk")) {
```
echeung-amzn, line 232: Perhaps have a separate `createDiskMetrics` function instead.

echeung-amzn (commented):
I'm not familiar with EC2/Nitro usage, so I'm not a good judge of what makes the most sense in this case. Do you have teammates that have any thoughts?

edisongustavo:
> I'm not familiar with EC2/Nitro usage, so I'm not a good judge of what makes the most sense in this case. Do you have teammates that have any thoughts?

From what I understood, these metrics just have a different prefix. I don't know why EC2 chose to do it this way, but that's the way it is. The metrics **without** the prefix don't appear for Nitro based instances, and vice-versa.

That's why I mentioned doing some metric math do "merge" both metrics so that the dashboard has a single time-series to look at.

I thought in something like this:

```
m1 = EBSDiskRead
m2 = DiskRead
AVG(REMOVE_EMPTY([m1, m2]))
```

I don't know if it would work, but maybe it's worth a shot.

test/monitoring/aws-ec2/EC2Monitoring.test.ts
edisongustavo: That's a good suggestion, I'll do that.

lib/monitoring/aws-ec2/EC2MetricFactory.ts
edisongustavo: That's a good suggestion, I'll do that.

test/monitoring/aws-ec2/EC2Monitoring.test.ts
edisongustavo: Do you mean not parametrizing the tests and have a single test instead?

test/monitoring/aws-ec2/EC2Monitoring.test.ts
edisongustavo: Not really because of how enums in Typescript work. I got this function from here: https://www.petermorlion.com/iterating-a-typescript-enum/.

I also found the [ts-enum-util](https://www.npmjs.com/package/ts-enum-util) library that handles this, which could be an alternative to hand-crafting my own function.

Another alternative would be to list the instance family types manually.

echeung-amzn: Ah you're right, I was thinking of string enums.

test/monitoring/aws-ec2/EC2Monitoring.test.ts
echeung-amzn: I mean doing something like

```ts
enumKeys(InstanceFamilyTypes).forEach(familyType => {
  test(...) {}
  
  test(...) {}
  
  test(...) {}
});
```

edisongustavo:
> Have you tried graphing metrics like that manually?

Yes, I've done that. It is as I said.

I am updating this PR with another commit which uses the MetricMath and in the end we have a single time series. This looks much nicer and doesn't require users of the library to change their code.

edisongustavo:
This is the final graph:

image

As you can see, there are no metrics for "classic" because the AutoScalingGroup is only for "nitro-based" instances.

echeung-amzn (approved):

PR #324, opened 2023-02-09
fix(rds): Clamp metric to 5-minutes period for DiskSpaceUsage

edisongustavo:
The `VolumeBytesUsed` metric has a minimum period of 5-minutes in my RDS clusters.

I didn't find this information documented in the AWS docs, but it gives hints for the other Volume* metrics here:
https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.AuroraMySQL.Monitoring.Metrics.html

At the monitoring of my RDS clusters I have changed to a 5-minute period, but I'd like to keep the 1-minute for the other metrics. We can discard this PR if you think this approach is not suitable.

---

### DIFF
--- a/lib/common/metric/MetricFactory.ts
+++ b/lib/common/metric/MetricFactory.ts
@@ -54,6 +54,11 @@ export interface MetricFactoryProps {
   readonly globalDefaults?: MetricFactoryDefaults;
 }
 
+export enum ResolvePeriodHandling {
+  ERROR,
+  CLAMP_TO_MINIMUM_REPORT_FREQUENCY,
+}
+
 export class MetricFactory {
   protected readonly globalDefaults: MetricFactoryDefaults;
 
@@ -94,7 +99,7 @@ export class MetricFactory {
         ? this.removeUndefinedEntries(dimensionsMap)
         : undefined,
       namespace: this.getNamespaceWithFallback(namespace),
-      period: period ?? this.globalDefaults.period ?? DefaultMetricPeriod,
+      period: this.resolvePeriod(period),
       region: region ?? this.globalDefaults.region,
       account: account ?? this.globalDefaults.account,
     });
@@ -121,7 +126,7 @@ export class MetricFactory {
       color,
       expression,
       usingMetrics,
-      period: period ?? this.globalDefaults.period ?? DefaultMetricPeriod,
+      period: this.resolvePeriod(period),
     });
   }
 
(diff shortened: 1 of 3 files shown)


lib/common/metric/MetricFactory.ts
```diff
@@ -435,6 +439,33 @@ export class MetricFactory {
     });
   }
 
+  resolvePeriod(
+    period?: Duration,
+    reportFrequency?: Duration,
+    metricName?: string,
+    errorHandling = ResolvePeriodHandling.ERROR
+  ): Duration {
+    let resolvedPeriod =
```
echeung-amzn, line 448: You could declare this as a `const` and just `return reportFrequency` directly when clamping.

lib/common/metric/MetricFactory.ts
```diff
@@ -54,6 +54,11 @@ export interface MetricFactoryProps {
   readonly globalDefaults?: MetricFactoryDefaults;
 }
 
+export enum ResolvePeriodHandling {
+  ERROR,
+  CLAMP_TO_MINIMUM_REPORT_FREQUENCY,
```
echeung-amzn, line 59: Can you add some docs to describe the behaviours a bit more?

That said, it seems safer to always error and force the implementer to be explicit with their intent on what the period should be.

lib/monitoring/aws-rds/RdsClusterMetricFactory.ts
```diff
@@ -59,21 +64,29 @@ export class RdsClusterMetricFactory {
     );
   }
 
-  metricFreeStorageInBytes() {
-    return this.metric("FreeLocalStorage", MetricStatistic.MIN, "Free");
+  metricFreeStorageInBytes(period?: Duration) {
```
echeung-amzn, line 67: Does this need to be clamped?

lib/monitoring/aws-rds/RdsClusterMetricFactory.ts
```diff
@@ -59,21 +64,29 @@ export class RdsClusterMetricFactory {
...
+    return this.metric("FreeLocalStorage", MetricStatistic.MIN, "Free", period);
   }
 
-  metricUsedStorageInBytes() {
-    return this.metric("VolumeBytesUsed", MetricStatistic.MAX, "Used");
+  metricUsedStorageInBytes(period?: Duration) {
```
echeung-amzn, line 71: Would like to see a test covering new behaviour.

echeung-amzn (changes requested):

edisongustavo:
I agree with your points. I tried to refactor the code but I think the code is getting too confusing for too little gain.

The issue that I think is the problem is that this is creating a new entity which is composed of

- desired period
- minimum period supported by metric

Then we'd have the handling of mismatched periods where we could either:

1. Throw an error if user specified a period not compatible with the minimum supported by the metric
2. Clamp the value
3. Ignore the mismatch

I will park this idea for now and maybe come back to it later. In the meantime I'm just using 5-minute periods in my RDS dashboard. If you disagree and think this is valuable, I can get back to it.

PR #376, opened 2023-06-04
feat(dynamo): add GSI throttling alarms

turanyuksel:
# Change Info
  GSI throttles can occur at times when the underlying table is not throttled. To notify GSI throttles, I added throttle alarms, adapting the pattern in `DynamoTableMonitoring`.

  # Testing Done
  Besides the package build, I deployed a variation of this change on an AWS stack.
 
  # Backwards Compatibility
  New `DynamoTableGlobalSecondaryIndexMonitoringProps` members are optional, so the change is backward-compatible.

---

### DIFF
--- a/lib/monitoring/aws-dynamo/DynamoTableGlobalSecondaryIndexMonitoring.ts
+++ b/lib/monitoring/aws-dynamo/DynamoTableGlobalSecondaryIndexMonitoring.ts
@@ -1,4 +1,8 @@
-import { GraphWidget, IWidget } from "aws-cdk-lib/aws-cloudwatch";
+import {
+  GraphWidget,
+  HorizontalAnnotation,
+  IWidget,
+} from "aws-cdk-lib/aws-cloudwatch";
 
 import {
   DynamoTableGlobalSecondaryIndexMetricFactory,
@@ -6,13 +10,16 @@ import {
 } from "./DynamoTableGlobalSecondaryIndexMetricFactory";
 import {
   BaseMonitoringProps,
+  CapacityType,
   CountAxisFromZero,
   DefaultGraphWidgetHeight,
   DefaultSummaryWidgetHeight,
+  DynamoAlarmFactory,
   MetricWithAlarmSupport,
   Monitoring,
   MonitoringScope,
   ThirdWidth,
+  ThrottledEventsThreshold,
 } from "../../common";
 import {
   MonitoringHeaderWidget,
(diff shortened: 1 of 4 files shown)


lib/monitoring/aws-dynamo/DynamoTableGlobalSecondaryIndexMonitoring.ts
```diff
@@ -74,6 +91,35 @@ export class DynamoTableGlobalSecondaryIndexMonitoring extends Monitoring {
...
+      const createdAlarm = this.gsiAlarmFactory.addThrottledEventsAlarm(
+        this.writeThrottleCountMetric,
+        CapacityType.WRITE,
+        alarmProps,
+        disambiguator
+      );
+      this.throttledEventsAnnotations.push(createdAlarm.annotation);
+      this.addAlarm(createdAlarm);
+    }
   }
```
echeung-amzn, line 123: Since there's alarms now, you'll need to add something like [this line](https://github.com/cdklabs/cdk-monitoring-constructs/blob/8999ae6cedc3648a82e9b31c0e78e342dbc2bdd7/lib/monitoring/aws-dynamo/DynamoTableMonitoring.ts#L332):

```ts
props.useCreatedAlarms?.consume(this.createdAlarms());
```

test/monitoring/aws-dynamo/DynamoTableGlobalSecondaryIndexMonitoring.test.ts
```diff
@@ -22,6 +22,16 @@ test("snapshot test: no alarms", () => {
...
+      Warning: {
+        maxThrottledEventsThreshold: 5,
+      },
+    },
+    addWriteThrottledEventsCountAlarm: {
+      Warning: {
+        maxThrottledEventsThreshold: 5,
+      },
+    },
   });
```
echeung-amzn, line 35: You'll be able to correctly assert that 2 alarms were created here too ([example](https://github.com/cdklabs/cdk-monitoring-constructs/blob/8999ae6cedc3648a82e9b31c0e78e342dbc2bdd7/test/monitoring/aws-dynamo/DynamoTableMonitoring.test.ts#L128)).

echeung-amzn (changes requested):

test/monitoring/aws-dynamo/DynamoTableGlobalSecondaryIndexMonitoring.test.ts
turanyuksel: Added another test case and the assertion.

echeung-amzn (approved):
Thanks for the contribution!

PR #405, opened 2023-08-09
feat(apigateway): allow using IRestApi instead of RestApiBase

bigbadtrumpet:
As of 2.90.0 aws-cdk now supports restApiName in the IRestApi interface. By using this interface it becomes possible for developers to monitor APIGateway without the need to reference the actual RestAPI.

Related commit: https://github.com/aws/aws-cdk/commit/3bf1361e20e2b2d497fcc2197fa45dac91e7eee3

---

### DIFF
--- a/lib/monitoring/aws-apigateway/ApiGatewayMetricFactory.ts
+++ b/lib/monitoring/aws-apigateway/ApiGatewayMetricFactory.ts
@@ -1,4 +1,4 @@
-import { RestApiBase } from "aws-cdk-lib/aws-apigateway";
+import { IRestApi } from "aws-cdk-lib/aws-apigateway";
 import { DimensionsMap } from "aws-cdk-lib/aws-cloudwatch";
 
 import {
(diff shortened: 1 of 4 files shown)


lib/monitoring/aws-apigateway/ApiGatewayMetricFactory.ts
```diff
@@ -14,9 +14,9 @@ const ApiGatewayNamespace = "AWS/ApiGateway";
 
 export interface ApiGatewayMetricFactoryProps {
   /**
-   * API to monitor (cannot use IRestApi, since it does not provide API name)
+   * API to monitor
    */
-  readonly api: RestApiBase;
+  readonly api: IRestApi;
```
echeung-amzn, line 19: Please add tests that use the interface in `ApiGatewayMonitoring.test.ts`.

echeung-amzn (changes requested):
Nice, I missed this. Looking at [the commit](https://github.com/aws/aws-cdk/commit/3bf1361e20e2b2d497fcc2197fa45dac91e7eee3), it seems to have been added as of `2.37.0` but we require `^2.65.0` so we're good on that front.

lib/monitoring/aws-apigateway/ApiGatewayMetricFactory.ts
bigbadtrumpet: Absolutely!

echeung-amzn (approved):

PR #411, opened 2023-08-15
feat(route53): Support Route53 Health Checks

marcogrcr:
WHAT?

Alarms created by `MonitoringFacade` can now be tagged with `PredefinedCustomTags.ROUTE_53_HEALTH_CHECK` to indicate and validate that the alarm will be used in a Route53 Health Check.

WHY?

Route53 Health Checks have strict requirements about which CloudWatch alarms can be used. Tagging an alarm with
`PredefinedCustomTags.ROUTE_53_HEALTH_CHECK` serves four purposes:

1. Documentation that the alarm is used in a Route53 Health Check.
2. Validation that the alarm will be able to be used in a Route53 Health Check.
3. Adjustment of the metric used to create the alarm to ensure it can be used in a Route53 Health Check.
4. Make it easy to find query the alarms that will be used to create Route53 Health Checks.

HOW?

`npx yarn build` succeeds all the way to `package.js`.

---


marcogrcr:
I debated internally whether `customTags` + `PredefinedCustomTags ` was the most appropriate way to implement this, or whether I should've added a new field to:

https://github.com/cdklabs/cdk-monitoring-constructs/blob/f23fd6736b90bbb601c87d7222b7438d9e8003fa/lib/common/alarm/CustomAlarmThreshold.ts#L12

I ultimately decided to go with the former because the field would also have to be added to:

https://github.com/cdklabs/cdk-monitoring-constructs/blob/f23fd6736b90bbb601c87d7222b7438d9e8003fa/lib/common/alarm/AlarmFactory.ts#L66

And it would've had to be populated on every `addSomeAlarm` method. While this would be mostly mitigated by the usage of the spread operator in the `lib/common/monitoring/alarms/` factory implementations, for example:

https://github.com/cdklabs/cdk-monitoring-constructs/blob/f23fd6736b90bbb601c87d7222b7438d9e8003fa/lib/common/monitoring/alarms/LatencyAlarmFactory.ts#L127-L144

Any user-defined subclass of `MonitoringFacade` that does not use the spread operator, would miss propagating the property accordingly. Additionally, using `customTags` makes it easy later to query all the appropriate alarms for creating the health checks as follows:

```ts
monitoring.createdAlarmsWithTag(PredefinedCustomTags.ROUTE_53_HEALTH_CHECK);
```

lib/common/alarm/AlarmFactory.ts
```diff
@@ -690,6 +704,111 @@ export class AlarmFactory {
...
+      adjustedMetric = adjustedMetric.with({ period: props.period });
+    }
+    if (adjustedMetric.label) {
+      // Annotations do not support dynamic labels, so we have to remove them from metric name
+      adjustedMetric = adjustedMetric.with({
+        label: removeBracketsWithDynamicLabels(adjustedMetric.label),
+      });
+    }
+
+    // Adjust metric for Route53 Health Checks
```
echeung-amzn, line 720: This logic should be abstracted outside of `AlarmFactory`; this is getting unwieldy.

echeung-amzn (commented):
> I debated internally whether customTags + PredefinedCustomTags  was the most appropriate way to implement this, or whether I should've added a new field to [CustomAlarmThreshold] [...] I ultimately decided to go with the former because the field would also have to be added to: [AddAlarmProps]

Maybe I'm missing something, but I don't think that's the case? The various `addSomeAlarm` methods should already be accepting interfaces that are extending `CustomAlarmThreshold`.

It feels like adding something like `metricRules` and having something like `ensureRoute53HealthCheckCompability()` that does the various validations/modifications, called from `adjustMetric`.

marcogrcr:
OK, I've taken your suggestion and added a new implementation. I've also updated https://github.com/cdklabs/cdk-monitoring-constructs/pull/411#issue-1850911095 to reflect the change. Hope this is in line with what you had in mind!

echeung-amzn (approved):

PR #415, opened 2023-08-28
feat: handle cross-account/region for metric math, search, and anomaly detection in MetricFactory

echeung-amzn:
This expands on what was implemented in #311, which only handled regular metrics via `createMetric`.

---


cayman-amzn (approved):
lgtm 🚢

Issue #418, opened 2023-08-30
KinesisDataStreamMonitoring read(andWrite)ProvisionedThroughputExceeded y axis shows up as ms


jeremy-zou:
### Version

3.0.0

### Steps and/or minimal code example to reproduce

1. Create AmazonMonitoringFacade object with createDashboard set to true
2. Create new or use existing Kinesis Stream
3. call Monitoring Facade object's `monitorKinesisDataStream()` with default params and Kinesis Stream from 2.
4. Inspect dashboard

### Expected behavior

The Provisioned Capacity Exceeded metric's y axis should be Count according to:

https://docs.aws.amazon.com/streams/latest/dev/monitoring-with-cloudwatch.html

### Actual behavior

The Provisioned Capacity Exceeded metric's y axis is ms

<img width="1395" alt="Screenshot 2023-08-30 at 12 04 09 PM" src="https://github.com/cdklabs/cdk-monitoring-constructs/assets/42885007/748e4756-11a6-4ccf-8e6b-429b4a276049">

PR #419, opened 2023-08-30
fix(kinesis): change Provisioned Capacity Exceeded metric's Y axis label to CountAxisFromZero

jeremy-zou:
Fixes https://github.com/cdklabs/cdk-monitoring-constructs/issues/418

---


echeung-amzn:
Thanks for the contribution! This also makes it clear that the naming of `metricReadProvisionedThroughputExceededPercent` and `metricWriteProvisionedThroughputExceededPercent` in `KinesisDataStreamMetricFactory` are incorrect.

Do you mind also:

1. Creating new methods there (`metricReadProvisionedThroughputExceeded` and `metricWriteProvisionedThroughputExceeded`)
2. Marking the existing methods as deprecated and just calling the new ones
3. Updating `KinesisDataStreamMonitoring` to call those new methods

We can remove the old methods later as a breaking change.

echeung-amzn (changes requested):
Looks like you're missing the changes to `API.md`. You should be able to run `yarn build` to produce the necessary changes.

jeremy-zou:
> Looks like you're missing the changes to `API.md`. You should be able to run `yarn build` to produce the necessary changes.

Ah yeah I see that. My bad; will fix.

echeung-amzn (approved):

PR #421, opened 2023-09-01
chore: add security policy doc [skip ci]

echeung-amzn:
Adding doc in standardized GitHub location so it's easily discoverable in the main project view.

---


cayman-amzn (changes requested):
Can we also update the README:

https://github.com/cdklabs/cdk-monitoring-constructs/blob/4bce1036e793083693bc4ba1390d8143b57072ca/README.md?plain=1#L418-L422

echeung-amzn:
Good catch, updated.

cayman-amzn (approved):

Issue #403, opened 2023-08-03
Metric Math Alarms do not adhere to minMetricSamplesToAlarm


akiesler:
### Version

5.3.4

### Steps and/or minimal code example to reproduce

1. Create a monitoring facade
2. Create a MathExpression using MetricFactory.createMetricMath()
3. Create an alarm using AlarmFactory.addAlarm()

### Expected behavior

I would expect the corresponding "NoSamples" alarm to either check each metrics has the necessary number of samples or wrap the entire expression in `DATAPOINT_COUNT` to check for the correct sample count.

### Actual behavior

The same metric alarm is created because there is no statistic prop to update.

akiesler:
For anyone else facing this same issue I was able to hack together this solution.
```typescript
import { Alarm, CfnAlarm, MathExpression } from 'aws-cdk-lib/aws-cloudwatch';
import { MonitoringFacade } from 'cdk-monitoring-constructs';

// Hack the underlying CFN types to be mutable so we can modify them.
type MetricDataQueryProperty = {
    -readonly [K in keyof CfnAlarm.MetricDataQueryProperty]: CfnAlarm.MetricDataQueryProperty[K];
};

type MetricStatProperty = {
    -readonly [K in keyof CfnAlarm.MetricStatProperty]: CfnAlarm.MetricStatProperty[K];
};

export function fixMetricMathMinSamples(monitoringFacade: MonitoringFacade) {
    // Find all minimum samples metric math alarms
    const alarms = monitoringFacade.node.children
        .filter((c) => c instanceof Alarm)
        .map((alarm) => alarm as Alarm)
        .filter((alarm) => alarm.node.id.includes('NoSamples') && alarm.metric instanceof MathExpression);

    alarms.forEach((alarm) => {
        const metrics = (alarm.node.defaultChild! as CfnAlarm).metrics! as MetricDataQueryProperty[];
        // Convert Metric Math to use the MAX Sample Count
        metrics.filter((m) => m.expression).forEach((m) => (m.expression = 'MAX(METRICS())'));
        // Convert Metrics to using SampleCount
        metrics.filter((m) => m.metricStat).forEach((m) => ((m.metricStat as MetricStatProperty).stat = 'SampleCount'));
    });
}
```
Would love to see the logic fixed in the library itself.

echeung-amzn:
[...]

PR #453, opened 2023-11-09
feat(alarm): add minSampleCountToEvaluateDatapoint

### LINKED ISSUE #452: minMetricSamplesToAlarm uses incorrect evaluation period
### Version

6.2.2

Reproducible with unit tests included in the package: https://github.com/cdklabs/cdk-monitoring-constructs/blob/84609a4d81ad5367ee14a442411c82fa4d368407/test/common/alarm/AlarmFactory.test.ts#L275

### Expected behavior

`NoSamples` child alarm should use the same `datapointsToAlarm` and `evaluationPeriods` as the alarm that we are attaching `minMetricSamplesToAlarm` to.

Example:

```
  factory.addAlarm(metric, {
    ...props,
    alarmNameSuffix: "none",
    comparisonOperator: ComparisonOperator.LESS_THAN_THRESHOLD,
    minMetricSamplesToAlarm: 500,
    datapointsToAlarm: 5,
    evaluationPeriods: 5,
  });
```

should create a `NoSamples` alarm with `datapointsToAlarm: 5` and ` evaluationPeriods: 5`.

### Actual behavior

`NoSamples` child alarm uses `datapointsToAlarm` and `evaluationPeriods` different than the main alarm - both are set `1` instead.

### Other details

The fact that the `datapointsToAlarm` and `evaluationPeriods`  for `NoSamples` child alarm are using different values (i.e. 1), causes incorrect Composite Alarm behaviour. 

Imagine monitor where you require 5 breaching datapoints to go into alarm with the `minMetricSamplesToAlarm` configured to an arbitrary value. Consider the following:

```
NoSamplesAlarm:         [X, X, X, X, 0, 0, 0, 0, 0]
MainAlarm:              [X, X, X, X, X, 0, 0, 0, 0]
---
CompositeWithSamples:   [0, 0, 0, 0, X, 0, 0, 0, 0]

where 0 is a non-breaching data point, X is a breaching data point
```

Even though there was only a single datapoint that satisfied both conditions, the monitor went into alarm despite the fact that the `MainAlarm` had `datapointsToAlarm` configured to 5 datapoints. `NoSamplesAlarm` generated by cdk-monitoring-constructs is based on `datapointsToAlarm: 1`, so if that single datapoint coincides with `MainAlarm` reaching its own threshold, the overall `CompositeWithSamples` will go into ALARM anyway. This is doesn't seem like a desirable behaviour - for the `CompositeWithSamples` to go into ALARM, all breaching datapoints should have sufficient number of samples.

If this is not a bug, is there any rationale behind the `NoSamples` using a different `datapointsToAlarm` and `evaluationPeriods`  than the `MainAlarm`?

miloszwatroba:
Fixes #452

Currently, when using `minMetricSamplesToAlarm` the number of samples is evaluated for a different period than the main alarm. This makes monitoring sensitive to false positives as not every breaching datapoint must have sufficient number of samples (see #452 for more details).

Moreover, the current approach for adjusting alarms to respect `minMetricSamplesToAlarm` is to create 2 extra alarms - one for `NoSamples` and one for a top-level composite. Each of these monitors incurs extra costs ($0.10 for `NoSamples` monitor and $0.50 for the Composite, see https://aws.amazon.com/cloudwatch/pricing/ for reference). This means that using `minMetricSamplesToAlarm` increases the cost from $0.10 per alarm to $0.70 per alarm ($0.60 of overhead!).

It's possible to use Math Expression instead. Instead of adding separate alarm for `NoSamples`, we can model it a Sample Count metric, and instead of the Composite, we can use the MathExpression that conditionally emits the data based on the number of samples. The charge for Math Expression-based alarms is per metric in the Math Expression, so that comes down to $0.20 per alarm. That's a 70% cost improvement. Additionally, it reduces the overall number of alarms, effectively making it easier to fit your alarming in the CloudWatch quota and decluttering the UI.

To avoid breaking any customers that rely on `minMetricSamplesToAlarm` generating alarms (e.g. https://github.com/cdklabs/cdk-monitoring-constructs/issues/403), deprecating it and adding `minSampleCountToEvaluateDatapoint` with updated behaviour next to it.

---

### DIFF
--- a/lib/common/alarm/AlarmFactory.ts
+++ b/lib/common/alarm/AlarmFactory.ts
@@ -7,6 +7,7 @@ import {
   CompositeAlarm,
   HorizontalAnnotation,
   IAlarmRule,
+  MathExpression,
   TreatMissingData,
 } from "aws-cdk-lib/aws-cloudwatch";
 import { Construct } from "constructs";
@@ -187,6 +188,18 @@ export interface AddAlarmProps {
   readonly evaluateLowSampleCountPercentile?: boolean;
 
   /**
+   * Specifies how many samples (N) of the metric is needed in a datapoint to be evaluated for alarming.
+   * If this property is specified, your metric will be subject to MathExpression that will add an IF condition
+   * to your metric to make sure that each datapoint is evaluated only if it has sufficient number of samples.
+   * If the number of samples is not sufficient, the datapoint will be treated as missing data and will be evaluated
+   * according to the treatMissingData parameter.
+   * If specified, deprecated minMetricSamplesToAlarm has no effect.
+   *
+   * @default - default behaviour - no condition on sample count will be used
+   */
+  readonly minSampleCountToEvaluateDatapoint?: number;
+
+  /**
    * Specifies how many samples (N) of the metric is needed to trigger the alarm.
    * If this property is specified, an artificial composite alarm is created of the following:
    * <ul>
@@ -195,6 +208,9 @@ export interface AddAlarmProps {
    * </ul>
    * The newly created composite alarm will be returned as a result, and it will take the original alarm actions.
    * @default - default behaviour - no condition on sample count will be added to the alarm
+   * @deprecated Use minMetricSampleCountToAlarm instead. minMetricSamplesAlarm uses different evaluation period
+   *   for its child alarms, so it doesn't guarantee that each datapoint in the evaluation period has sufficient
+   *   number of samples
    */
   readonly minMetricSamplesToAlarm?: number;
 
@@ -511,6 +527,9 @@ export class AlarmFactory {
       props
     );
 
+    // metric that will be ultimately used to create the alarm
+    let alarmMetric: MetricWithAlarmSupport = adjustedMetric;
+
     // prepare primary alarm properties
 
     const actionsEnabled = this.determineActionsEnabled(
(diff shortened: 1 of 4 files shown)


lib/common/alarm/AlarmFactory.ts
```diff
@@ -549,32 +568,56 @@ export class AlarmFactory {
...
+          sampleCount: metricSampleCount,
+        },
+      });
+    }
+
     // create primary alarm
 
-    const primaryAlarm = adjustedMetric.createAlarm(
-      this.alarmScope,
+    const primaryAlarm = alarmMetric.createAlarm(this.alarmScope, alarmName, {
```
miloszwatroba, line 597: adjustedMetric changed to alarmMetric; the rest was automatically reformated by `yarn build` without any changes

lib/common/alarm/AlarmFactory.ts
```diff
@@ -195,6 +208,9 @@ export interface AddAlarmProps {
    * </ul>
    * The newly created composite alarm will be returned as a result, and it will take the original alarm actions.
    * @default - default behaviour - no condition on sample count will be added to the alarm
+   * @deprecated Use minMetricSampleCountToAlarm instead. minMetricSamplesAlarm uses different evaluation period
```
xendo, line 211: you mean `minSampleCountToEvaluateDatapoint` ?

lib/common/alarm/AlarmFactory.ts
```diff
@@ -549,32 +568,56 @@ export class AlarmFactory {
       );
     }
 
+    // apply metric math for minimum metric samples
+
+    if (props.minSampleCountToEvaluateDatapoint) {
+      if (adjustedMetric instanceof MathExpression) {
+        throw new Error(
+          "minSampleCountToEvaluateDatapoint is not supported for MathExpressions"
```
xendo, line 576: I would add a sentence that if you already use metric math you can extend your expression to do the sample count evaluation.

xendo (commented):
LGTM, two nitpicks

lib/common/alarm/AlarmFactory.ts
miloszwatroba: good idea, done

lib/common/alarm/AlarmFactory.ts
```diff
@@ -549,32 +568,58 @@ export class AlarmFactory {
...
     }
 
+    // apply metric math for minimum metric samples
+
+    if (props.minSampleCountToEvaluateDatapoint) {
+      if (adjustedMetric instanceof MathExpression) {
+        throw new Error(
+          "minSampleCountToEvaluateDatapoint is not supported for MathExpressions. " +
+            "If you already use MathExpression, you can extend your expression to evaluate " +
+            "the sample count using IF statement, e.g. IF(sampleCount > X, mathExpression)."
```
echeung-amzn, line 578: I wonder if it's worth abstracting that at some point anyway just for convenience?

echeung-amzn (approved):

PR #471, opened 2023-12-15
feat(ecs-patterns): allow imported IBaseService

echeung-amzn:
The existing logic to get metrics on `BaseService` [is pretty simple](https://github.com/aws/aws-cdk/blob/v2.115.0/packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts#L1195-L1220) and can probably be lifted up to `IBaseService` upstream (similar [to the change for load balancers](https://github.com/aws/aws-cdk/commit/cb889bc2c267654ca97e3d85a16a99a667d3584c)), but I don't have bandwidth to tackle that right now.

---


lib/monitoring/aws-ecs-patterns/BaseServiceMetricFactory.ts
```diff
@@ -1,44 +1,60 @@
...
   }
 
   metricClusterCpuUtilisationInPercent() {
-    return this.metricFactory.adaptMetric(
-      this.service.metricCpuUtilization({
-        label: "Cluster CPU Utilization",
-      })
+    return this.metricFactory.createMetric(
+      "CPUUtilization",
+      MetricStatistic.AVERAGE,
```
ayush987goyal, line 42: Did the metricFactory used to adapt this metric to `Average` stats before? If not, then would it be a breaking change in regards to the default stats of this metric?

echeung-amzn: It's just a regular `Metric` ([code ref](https://github.com/aws/aws-cdk/blob/166967f11727a28fc11b9af5de0fad6da2a4ad64/packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts#L1222)), which defaults to `Average` ([code ref](https://github.com/aws/aws-cdk/blob/v2.117.0/packages/aws-cdk-lib/aws-cloudwatch/lib/metric.ts#L44)).

PR #477, opened 2024-01-05
feat: Keep NLB and ALB information separated to avoid check of constructor name

### LINKED ISSUE #476: Cannot use NetworkLoadBalancer.fromNetworkLoadBalancerAttributes with .monitorFargateNetworkLoadBalancer
### Version

"7.2.2"

1. Init Facade
2. call `monitoring.monitorFargateNetworkLoadBalancer({ networkLoadBalancer:  NetworkLoadBalancer.fromNetworkLoadBalancerAttributes(...) })`

You will get `Invalid type of load balancer or target group (only ALB and NLB are supported).`

This is happening that internally code is too much shared with ALB and it tries to get type from name of constructor function/class, which in case of imported NLB will not contain Network in name.

### Expected behavior

It should work with imported NLB/ALB

### Actual behavior

Throwing error `Invalid type of load balancer or target group (only ALB and NLB are supported).`

### Other details

In each function you ask for specific type of ALB/NLB via separate properties `networkLoadBalancer` and `applicationLoadBalancer` - so you can keep this information internally

btd:
Fixes #476

---

I did not updated EC2 staff, want first get some feedback


echeung-amzn:
Hi, thanks for the issue report and PR! An issue with this proposed change is that it'd be a breaking API change due to the prop changes.

I've opened another PR that might be a better alternative to this: #480

Let me know if you have any thoughts!

btd:
@echeung-amzn thanks. Your version is quite hacky, but yes it is not BC. So i am all for it as upgrade process will be more simple

PR #480, opened 2024-01-16
feat(ecs-patterns): improve ALB/NLB type guards

### LINKED ISSUE #476: Cannot use NetworkLoadBalancer.fromNetworkLoadBalancerAttributes with .monitorFargateNetworkLoadBalancer

echeung-amzn:
Fixes #476

This is an alternative to PR #477 without a breaking API change.

---

### DIFF
--- a/lib/monitoring/aws-loadbalancing/LoadBalancerMetricFactory.ts
+++ b/lib/monitoring/aws-loadbalancing/LoadBalancerMetricFactory.ts
@@ -12,29 +12,27 @@ import { MetricFactory, MetricWithAlarmSupport } from "../../common";
 // It's not possible to use instanceOf with typescript interfaces, so we have to use type guards to differentiate
 // between the interfaces. As another problem, the LoadBalancer/TargetGroup for both Network and Application types
 // don't have distinguished fields that could be used to differentiate between both types, so we resort to using
-// the name of the class below.
-//
-// Ideally the 2 interfaces would provide a specific field to distinguish both types.
+// checking for unique methods in their metrics interfaces.
 function isApplicationLoadBalancer(
   loadBalancer: INetworkLoadBalancer | IApplicationLoadBalancer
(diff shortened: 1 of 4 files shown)


cayman-amzn (approved):

lib/monitoring/aws-loadbalancing/LoadBalancerMetricFactory.ts
```diff
@@ -12,29 +12,27 @@ import { MetricFactory, MetricWithAlarmSupport } from "../../common";
 // It's not possible to use instanceOf with typescript interfaces, so we have to use type guards to differentiate
 // between the interfaces. As another problem, the LoadBalancer/TargetGroup for both Network and Application types
 // don't have distinguished fields that could be used to differentiate between both types, so we resort to using
-// the name of the class below.
-//
-// Ideally the 2 interfaces would provide a specific field to distinguish both types.
+// checking for unique methods in their metrics interfaces.
```
ayush987goyal, line 15: Checking on unique metrics methods still feels iffy but I guess that's all we have right now.

ayush987goyal (approved):

PR #496, opened 2024-02-07
feat(rds): add support for RDS Instance Monitoring

### LINKED ISSUE #236: [rds] Add Monitoring for RDS DB Instance
### Feature scope

RDS

The library currently supports out-of-the-box monitoring for RDS Clusters using [monitorRdsCluster](https://github.com/cdklabs/cdk-monitoring-constructs/blob/a371e31593fa3ef96a7dd74a47937f2cf2537395/lib/facade/MonitoringFacade.ts#L589-L593)

Would it be possible to add support to monitor [RDS DB Instances](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.DatabaseInstance.html)? We currently have to roll our own widgets to be able to do this.

ayvazj:
Fixes #236 

Added support for monitoring RDS Instances. Implementation is largely based on RDS Cluster monitoring code and includes the metrics necessary to recreate the "automatic" Cloud Watch dashboard. This implementation supports my needs, but further metrics can be added if there is a interest.

---


lib/facade/IMonitoringAspect.ts
```diff
@@ -63,7 +64,9 @@ export interface MonitoringAspectProps {
   readonly kinesisFirehose?: MonitoringAspectType<KinesisFirehoseMonitoringOptions>;
   readonly lambda?: MonitoringAspectType<LambdaFunctionMonitoringOptions>;
   readonly openSearch?: MonitoringAspectType<OpenSearchClusterMonitoringOptions>;
-  readonly rds?: MonitoringAspectType<RdsClusterMonitoringOptions>;
+  readonly rds?: MonitoringAspectType<
+    RdsClusterMonitoringOptions | RdsInstanceMonitoringOptions
```
echeung-amzn, line 68: This would probably make it hard to specify options for both if you wanted to. How about a separate one for `rdsInstance`?

We could also add a `rdsCluster` that takes precedence over `rds` and mark `rds` as deprecated if the naming seems confusing.

lib/monitoring/aws-rds/RdsInstanceMetricFactory.ts
```diff
@@ -0,0 +1,139 @@
...
+    this.instanceIdentifier =
+      RdsInstanceMetricFactory.resolveDbInstanceIdentifier(props);
+    this.dimensionsMap = { DBInstanceIdentifier: this.instanceIdentifier };
+  }
+
+  private static resolveDbInstanceIdentifier(
+    props: RdsInstanceMetricFactoryProps
+  ): string {
+    if (!props.instance) {
+      throw Error("`instance` is required");
```
echeung-amzn, line 42: Why allow it to be nullable in the props then?

lib/monitoring/aws-rds/RdsInstanceMetricFactory.ts
```diff
@@ -0,0 +1,139 @@
...
+  metricTotalConnectionCount() {
+    return this.metric(
+      "DatabaseConnections",
+      MetricStatistic.SUM,
+      "Connections: Sum"
+    );
+  }
+
+  metricAverageCpuUsageInPercent() {
+    return this.metric("CPUUtilization", MetricStatistic.AVERAGE, "CPU Usage");
```
echeung-amzn, line 56: We could just use the helper functions like https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.IDatabaseInstance.html#metricwbrcpuutilizationprops similar to [this](https://github.com/cdklabs/cdk-monitoring-constructs/blob/f59da8e8061d625eb7ae9b622512479a715a146f/lib/monitoring/aws-sns/SnsTopicMetricFactory.ts#L19-L22).

echeung-amzn (changes requested):
In general looking pretty good!

lib/facade/MonitoringAspect.ts
```diff
@@ -320,6 +320,12 @@ export class MonitoringAspect implements IAspect {
         alarmFriendlyName: node.node.path,
         ...props,
       });
+    } else if (isEnabled && node instanceof rds.DatabaseInstance) {
```
echeung-amzn, line 323: Similar to the splitting of `rds` to `rdsCluster`/`rdsInstance`, make `monitorRds` into 2 methods too using the updated props.

echeung-amzn (changes requested):

lib/facade/MonitoringAspect.ts
```diff
@@ -312,8 +313,8 @@ export class MonitoringAspect implements IAspect {
     }
   }
 
-  private monitorRds(node: IConstruct) {
-    const [isEnabled, props] = this.getMonitoringDetails(this.props.rds);
+  private monitorRdsCluster(node: IConstruct) {
+    const [isEnabled, props] = this.getMonitoringDetails(this.props.rdsCluster);
```
echeung-amzn, line 317: Since the `rds` prop is only deprecated and not removed yet, we need to actually fall back to it to retain backwards compatibility with something like:

```
this.props.rdsCluster ?? this.props.rds
```

echeung-amzn (changes requested):

echeung-amzn (approved):

PR #519, opened 2024-05-10
fix(AlarmFactory): Fix wording for actionsSuppressorExtensionPeriod prop to use 'duration' instead of 'time in seconds'

jonathan-luo:
### Description
No code logic changes.

Unfortunately missed changing `actionsSuppressorExtensionPeriod` prop description to use "duration" instead of "time in seconds". The "time in seconds" descriptor is used in [CloudFormation documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudwatch-compositealarm.html), but since the prop type required by CDK `CompositeAlarm` is `Duration`, it's more accurate to use "duration" (and to be consistent with the [CDK Composite Alarm definition](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudwatch.CompositeAlarm.html)).

Note: This wording change was already done for `actionsSuppressorWaitPeriod`; only `actionsSuppressorExtensionPeriod` was missed.

### Testing
- `yarn build` -- no code logic changes so no additional unit tests necessary.

---


jonathan-luo:
Initial build failed since updated `API.md` was not included in commit. Have fixed that now.

echeung-amzn (approved):

PR #533, opened 2024-06-25
feat: allow specifying regions/accounts at a more granular level

echeung-amzn:
This allows users to target different regions/account for specific things within a facade rather than relying on 1 facade per region/account with the facade's global settings, so something like:

```ts
this.facade
  .monitorDynamoTable({
    ...,
  })
  .monitorDynamoTable({
    ...,
    region: <some_other_region>,
    account: <some_other_account>,
  })
```

---

### DIFF
--- a/lib/common/metric/BaseMetricFactory.ts
+++ b/lib/common/metric/BaseMetricFactory.ts
@@ -1,15 +1,33 @@
 import { MetricFactory } from "./MetricFactory";
 
 export interface BaseMetricFactoryProps {
-  // TODO: this will eventually include other common things like account/region
+  /**
+   * Region where the metrics exist.
+   *
+   * @default The region configured by the construct holding the Monitoring construct
+   * @see https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Cross-Account-Cross-Region.html
+   */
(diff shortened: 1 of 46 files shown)


lib/common/metric/BaseMetricFactory.ts
```diff
@@ -1,15 +1,33 @@
 import { MetricFactory } from "./MetricFactory";
 
 export interface BaseMetricFactoryProps {
-  // TODO: this will eventually include other common things like account/region
+  /**
+   * Region where the metrics exist.
+   *
+   * @default The region configured by the construct holding the Monitoring construct
```
ayush987goyal, line 7: Will this default be overridden by the one defined on the facade? We should probably mention the order (segment > facade > construct)?

lib/common/metric/MetricFactory.ts
```diff
@@ -31,21 +31,6 @@ export interface MetricFactoryDefaults extends BaseMetricFactoryProps {
    * @default - DefaultMetricPeriod
    */
   readonly period?: Duration;
-
```
ayush987goyal, line 34: Is this a breaking change?

ayush987goyal (approved):

PR #537, opened 2024-06-26
feat(loadbalancing): add fail-open monitoring

### LINKED ISSUE #532: Add fail-open metrics to Network Load Balancer monitoring
### Feature scope

AWS/NetworkELB

Hi,

I would like to propose adding a widget for fail-open monitoring in AWS Network Load Balancer monitoring. This will give a better visibility into whether the fail-open mechanism was triggered or not whenever all targets are unhealthy.

Documentation: https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-cloudwatch-metrics.html

Metric name: `UnhealthyRoutingFlowCount`
Available dimensions: 
* `LoadBalancer`
* `AvailabilityZone, LoadBalancer`

Is this something that we would like to integrate to the NLB monitoring constructs? If so, happy to contribute.

miloszwatroba:
Fixes #532

This pull request adds fail-open monitoring to our Load Balancing monitors.

To give context about fail open ([ref](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html)):
> If a target group contains only unhealthy registered targets, the load balancer routes requests to all those targets, regardless of their health status. This means that if all targets fail health checks at the same time in all enabled Availability Zones, the load balancer fails open. The effect of the fail-open is to allow traffic to all targets in all enabled Availability Zones, regardless of their health status, based on the load balancing algorithm

Adding this metric will give better visibility into whether the Load Balancer's fail open routing was used during incidents. The metrics added as part of this pull request are in line with the AWS documentation:
* `UnhealthyRoutingRequestCount` for ApplicationLoadBalancer with `LoadBalancer` and `TargetGroup` dimensions ([ref](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-cloudwatch-metrics.html))
* `UnhealthyRoutingFlowCount` for NetworkLoadBalancer with `LoadBalancer` dimension ([ref](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-cloudwatch-metrics.html))

These metrics are reported conditionally, only when they have nonzero values. Thus, I added a `FILL(metric, 0)` metric math to correctly represent the values on the dashboards.

Tested with NetworkLoadBalancer, as that's the setup I have on my account - the ApplicationLoadBalancer values are based only on the documentation.

**Discussion point**: I added `metricUnhealthyRoutingCount` to the `ILoadBalancerMetricFactory` interface, which is exported from the package. This means, that this commit theoretically introduced a breaking change for customers that extended the library with their own classes that implement this interface. How do you approach such changes in this library Should this commit be considered a breaking change and released in a new major version or is this overly cautious? 

---

### DIFF
--- a/API.md
+++ b/API.md
@@ -48951,6 +48951,7 @@ new ApplicationLoadBalancerMetricFactory(metricFactory: MetricFactory, props: Ap
 | <code><a href="#cdk-monitoring-constructs.ApplicationLoadBalancerMetricFactory.metricHealthyTaskInPercent">metricHealthyTaskInPercent</a></code> | *No description.* |
 | <code><a href="#cdk-monitoring-constructs.ApplicationLoadBalancerMetricFactory.metricNewConnectionCount">metricNewConnectionCount</a></code> | *No description.* |
 | <code><a href="#cdk-monitoring-constructs.ApplicationLoadBalancerMetricFactory.metricProcessedBytesMin">metricProcessedBytesMin</a></code> | *No description.* |
+| <code><a href="#cdk-monitoring-constructs.ApplicationLoadBalancerMetricFactory.metricUnhealthyRoutingCount">metricUnhealthyRoutingCount</a></code> | *No description.* |
 | <code><a href="#cdk-monitoring-constructs.ApplicationLoadBalancerMetricFactory.metricUnhealthyTaskCount">metricUnhealthyTaskCount</a></code> | *No description.* |
 
 ---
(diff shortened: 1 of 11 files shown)


echeung-amzn:
> Should this commit be considered a breaking change and released in a new major version or is this overly cautious?

A quick skim over internal codebases and [a GitHub code search](https://github.com/search?q=ILoadBalancerMetricFactory&type=code) doesn't seem to reveal any such usages, so I'd consider it relatively safe. 

If it turns out I'm wrong and you've ended up here, please open up an issue if this is a problem for your setup.

echeung-amzn (approved):

PR #544, opened 2024-07-12
fix: apply account/region to more metrics

### LINKED ISSUE #428: Cannot create an Alarm based on a MathExpression which specifies a searchAccount or searchRegion
### Version

5.10.0

I have a few constructs that create DDB alarms, they used to be work with 5.4.2 but fails to build with 5.10.0, I am getting

```
Cannot create an Alarm based on a MathExpression which specifies a searchAccount or searchRegion
```

Here is my alarm config

```
addReadThrottledEventsCountAlarm: {
  Critical: {
    maxThrottledEventsThreshold: 0,
    datapointsToAlarm: 1,
    period: Duration.minutes(1),
    treatMissingDataOverride: TreatMissingData.NOT_BREACHING,
  }
}
```

### Expected behavior

DDB alarms are created.

### Actual behavior

Not able to get alarms created.

### Other details

I think this is related to #415 where we added `searchAccount` and `searchRegion` to `createMetricMath` and ReadThrottledEventsCountAlarm and WriteThrottledEventsCountAlarm use math expression.

https://github.com/cdklabs/cdk-monitoring-constructs/blob/b07f3615de87e160081834d4816a1617565a1e76/lib/monitoring/aws-dynamo/DynamoTableMetricFactory.ts#L138-L141

The change impacts more than just DDB alarms I feel, it affects all alarm creations where math expression will be used

### LINKED ISSUE #434: MetricFactory `toRate` does not pass ".with(properties)" to underlying metric
### Version

5.0.0

1. Create a MetricFactory with a tps metric.
2. use .with(properties); on the tps metric
3. Observe that properties aren't included on the tps metric itself.

Relevant line of code:
https://github.com/cdklabs/cdk-monitoring-constructs/blob/064d54fca62f7f62ed68ef40870db672cc7ee9f5/lib/monitoring/aws-apigateway/ApiGatewayMetricFactory.ts#L77

### Expected behavior

I would expect that properties are passed to the underlying metric.  This functionality works on metrics that do not use the "toRate" method as the metric is directly returned.

For example, .with(properties) works correctly on metric4XXErrorCount: https://github.com/cdklabs/cdk-monitoring-constructs/blob/064d54fca62f7f62ed68ef40870db672cc7ee9f5/lib/monitoring/aws-apigateway/ApiGatewayMetricFactory.ts#L98

### Actual behavior

Properties are not passed to underlying TPS metric.

### Other details

We are using this to show a widget for TPS, but need to specify the account.  The account for the TPS metric we specify in the properties is not being used.

echeung-amzn:
Addressing a few misses from #533.

Also updates the test in `MonitoringFacade.test.ts` since XAXR alarms aren't possible.

Fixes #434, but only in cases where we can infer that the account/region is actually different from the construct scope to prevent unnecessary XAXR alarm errors.

Fixes #428, in cases where the account/region is actually different from the construct scope.

---


lib/common/metric/MetricFactory.ts
```diff
@@ -452,4 +476,58 @@ export class MetricFactory {
...
+    if (!this.scope) {
+      return metricAccount;
+    }
+
+    const { account } = Stack.of(this.scope);
+    if (metricAccount !== account) {
+      return metricAccount;
+    }
+
+    return;
```
ayush987goyal, line 495: What happens if the `metricAccount` is not defined? Do we want to return `account` in those cases or we default to not passing account (which resolves to same account)?

Same question for region below.

echeung-amzn: If they're not explicitly declared, it should just inherit what the deployment environment context is (i.e. whatever the stack is in, like what a `Metric` would normally do if you don't pass in `account` or `region`).

ayush987goyal (approved):

PR #562, opened 2024-08-20
fix(stepfunctions): specify line colors for Step Functions Executions graph to ensure failed executions don't show as green

kamalgill:
The Step Functions Executions graph doesn't specify line colors for the various metrics, which may lead to failed executions displaying as a green line, and succeeded executions displaying as an orange line, as in the screenshot below.  This PR specifies line colors to avoid that anomaly. 

Screenshot 2024-08-20 at 11 00 05 AM

---


adamstein-eb (commented):
LGTM

echeung-amzn (changes requested):
We generally omitted these and left it to CloudWatch defaults so we didn't have to handle a11y concerns too.

The motivation for this is understandable though, but can you at least use (/update) the colors in https://github.com/cdklabs/cdk-monitoring-constructs/blob/main/lib/common/widget/color.ts so they can be consistent across other usages, if any?

kamalgill:
@echeung-amzn I have pushed a follow-up commit based on your feedback

echeung-amzn (changes requested):
Looks like the build failed. You'll need to run a full build locally (i.e. `yarn build`) and commit any updated snapshot files.

kamalgill:
@echeung-amzn I have updated the snaphots via `yarn build` and have squashed my commits in this PR

echeung-amzn (approved):

PR #564, opened 2024-08-26
Update Statistic for DatabaseConnection metric in RDS

veaditya:
As per AWS documentation, DatabaseConnection should use sample count statistic for DatabaseConnection metric. 

AWS Documentation: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-metrics.html

Fixes #

---


echeung-amzn:
> As per AWS documentation, for RDS DatabaseConnection metric, SampleCount statistic should be used.

I only see on the linked page that the metric is of unit `Count`. Why would using sample count make sense?

veaditya:
I confused Count with SampleCount. Apologies. I will cancel this pull request.

PR #569, opened 2024-09-09
feat(lambda): add isIterator prop

echeung-amzn:
We've received feedback that the default IteratorAge widget is largely useless for any Lambda function that isn't actually a stream handler. Function doesn't provide a way to really tell if that's the case, so we ended up with an explicit prop.

We default to true for backwards compat.

---

### DIFF
--- a/lib/monitoring/aws-lambda/LambdaFunctionMonitoring.ts
+++ b/lib/monitoring/aws-lambda/LambdaFunctionMonitoring.ts
@@ -23,6 +23,7 @@ import {
   ErrorCountThreshold,
   ErrorRateThreshold,
   ErrorType,
+  HalfWidth,
   HighTpsThreshold,
   LatencyAlarmFactory,
   LatencyThreshold,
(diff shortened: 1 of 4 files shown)


lib/monitoring/aws-lambda/LambdaFunctionMonitoring.ts
```diff
@@ -493,6 +506,12 @@ export class LambdaFunctionMonitoring extends Monitoring {
       this.addAlarm(createdAlarm);
     }
     for (const disambiguator in props.addMaxIteratorAgeAlarm) {
+      if (!this.isIterator) {
+        throw new Error(
+          "addMaxIteratorAgeAlarm is not applicable if isIterator is not true",
+        );
```
cayman-amzn, line 512: I feel I've asked this before, so sorry in advance if i have

Can we use `Annotations` here? Having the scope included (via `Annotations.error(...)`) really helps folks find where in their app this is being thrown

cayman-amzn (approved):

PR #591, opened 2024-11-08
fix(MetricFactory): wrap search query dimensions with quotes to handle potential whitespace

echeung-amzn:
---


giantcow (approved):

dgmezan-amzn:
The namespace fix doesn't work in case where we need a multiple namespace. For example,

## Without this commit
{my_namespace,reason}

## With this commit
{"my_namespace,reason"}

## Real fix for space namespaces
{"my_namespace","reason"}

PR #594, opened 2024-11-12
feat!: v9 breaking changes

echeung-amzn:
**DO NOT MERGE AS-IS; I need to merge this non-squashed to retain a sensible commit history.**

This PR includes some breaking changes to the API and minimum `aws-cdk-lib` peer dependency to unblock usage of newer features.

Refer to the individual commits for more details.

---

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license_

PR #596, opened 2024-11-13
chore: remove explicit @types/prettier override

echeung-amzn:
Extracting from #594 since it wasn't really relevant.

The problem with the types being incompatible has been long resolved.

---


echeung-amzn:
No, it should be fine to pick up newer versions by now.

PR #621, opened 2025-02-11
feat(lambda): allow latency alarms to be based on function timeout

### LINKED ISSUE #567: `monitorScope` always create latency alarm for lambda based on timeout value
### Feature scope

Lambda

The latency alarms can be added to lambdas on static value at the moment.
It would be great to set this to alarm on percentage of the lambda timeout value.
This is more important when we want to monitor scope and create alarm for all lambdas within the scope using `monitorScope`, and each lambda has it own timeout and not one duration value fits all.

AminFazlMondo:
Fixes #567 

---

### DIFF
--- a/lib/common/monitoring/alarms/LatencyAlarmFactory.ts
+++ b/lib/common/monitoring/alarms/LatencyAlarmFactory.ts
@@ -136,6 +136,11 @@ export interface LatencyThreshold extends CustomAlarmThreshold {
   readonly maxLatency: Duration;
 }
 
+export interface LatencyTimeoutPercentageThreshold
+  extends CustomAlarmThreshold {
+  readonly maxLatencyPercentage: number;
+}
+
 export interface DurationThreshold extends CustomAlarmThreshold {
   readonly maxDuration: Duration;
 }
--- a/lib/common/monitoring/alarms/LatencyAlarmFactory.ts
+++ b/lib/common/monitoring/alarms/LatencyAlarmFactory.ts
@@ -138,7 +138,7 @@ export interface LatencyThreshold extends CustomAlarmThreshold {
 
 export interface LatencyTimeoutPercentageThreshold
   extends CustomAlarmThreshold {
-  readonly maxLatencyPercentage: number;
+  readonly maxLatencyPercentageOfTimeout: number;
 }
 
 export interface DurationThreshold extends CustomAlarmThreshold {
--- a/lib/monitoring/aws-lambda/LambdaFunctionMonitoring.ts
+++ b/lib/monitoring/aws-lambda/LambdaFunctionMonitoring.ts
@@ -402,7 +402,6 @@ export class LambdaFunctionMonitoring extends Monitoring {
       this.latencyAnnotations.push(createdAlarm.annotation);
       this.addAlarm(createdAlarm);
     }
-    (props.lambdaFunction as Function).timeout;
     for (const disambiguator in props.addLatencyP99Alarm) {
       const alarmProps = props.addLatencyP99Alarm[disambiguator];
       const createdAlarm = this.latencyAlarmFactory.addLatencyAlarm(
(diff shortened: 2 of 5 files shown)


AminFazlMondo:
Hi @echeung-amzn ,

Could you please have a look at this?

Cheers,

lib/monitoring/aws-lambda/LambdaFunctionMonitoring.ts
```diff
@@ -558,6 +572,24 @@ export class LambdaFunctionMonitoring extends Monitoring {
...
+  private convertToLatencyThreshold(
+    latencyProps: LatencyThreshold | LatencyTimeoutPercentageThreshold,
+    lambdaFunction: IFunction,
+  ): LatencyThreshold {
+    if ("maxLatency" in latencyProps) {
+      return latencyProps;
+    }
+    const { maxLatencyPercentage, ...rest } = latencyProps;
+    const timeout =
+      (lambdaFunction as Function).timeout ?? Duration.seconds(30);
```
echeung-amzn, line 584: The default is 3 seconds.

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.Function.html#timeout

lib/monitoring/aws-lambda/LambdaFunctionMonitoring.ts
```diff
@@ -383,18 +396,19 @@ export class LambdaFunctionMonitoring extends Monitoring {
...
         this.p90LatencyMetric,
         LatencyType.P90,
-        alarmProps,
+        this.convertToLatencyThreshold(alarmProps, props.lambdaFunction),
         disambiguator,
       );
       this.latencyAnnotations.push(createdAlarm.annotation);
       this.addAlarm(createdAlarm);
     }
+    (props.lambdaFunction as Function).timeout;
```
echeung-amzn, line 405: This isn't used, can remove.

lib/common/monitoring/alarms/LatencyAlarmFactory.ts
```diff
@@ -136,6 +136,11 @@ export interface LatencyThreshold extends CustomAlarmThreshold {
   readonly maxLatency: Duration;
 }
 
+export interface LatencyTimeoutPercentageThreshold
+  extends CustomAlarmThreshold {
+  readonly maxLatencyPercentage: number;
```
echeung-amzn, line 141: By itself without seeing the interface name, it's unclear what this actually means.

Maybe `maxLatencyPercentageOfTimeout`?

echeung-amzn (changes requested):
1. There's no tests?
2. I feel pretty iffy about this, it feels like a janky workaround just because you're relying on `monitorScope`, where it feels more like you shouldn't be relying on that for such more complex use cases if you actually want fine-grained control.

AminFazlMondo:
> 1. There's no tests?
> 2. I feel pretty iffy about this, it feels like a janky workaround just because you're relying on `monitorScope`, where it feels more like you shouldn't be relying on that for such more complex use cases if you actually want fine-grained control.

Thanks @echeung-amzn  for getting back to me on this,
I have made the changes as you requested, also have added some tests.
I think the latency alarms do actually relate to the timeout of the lambda and this would allow for not repeating these values either if you are monitoring a single lambda or the whole scope of your stack.
Cheers,

test/monitoring/aws-lambda/LambdaFunctionMonitoring.test.ts
```diff
@@ -629,3 +629,110 @@ test("doesn't create alarms for enhanced Lambda Insights metrics if not enabled"
 
   expect(numAlarmsCreated).toStrictEqual(0);
 });
+
+test("snapshot test: latency alarms with percentage of timeout with default value", () => {
+  const stack = new Stack();
+
+  const scope = new TestMonitoringScope(stack, "Scope");
+
+  const lambdaFunction = new Function(stack, "Function", {
```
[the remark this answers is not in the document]
AminFazlMondo: updated

echeung-amzn (changes requested):
Build's failing.

AminFazlMondo:
> Build's failing.

Updated now, should be OK.

PR #627, opened 2025-03-24
feat(KinesisFirehoseMonitoring): add isDataFormatConversionEnabled prop

pbsinclair42:
The conversions widget is useless for any Firehose which doesn't have data format conversions enabled.

Default is true for backwards compatability.

Implementation is based on the similar `isIterator` parameter for Lambda Functions. 

Note that the AWS docs are not consistent about what they call this feature, [sometimes calling it data format conversion, sometimes calling it record format conversion](https://docs.aws.amazon.com/firehose/latest/dev/enable-record-format-conversion.html). CDK calls it [data format conversion](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_kinesisfirehose.CfnDeliveryStream.DataFormatConversionConfigurationProperty.html), hence my choice to use that terminology for the parameter too.

The library doesn't currently support enabling alarms on any of the data format conversion metrics, so we don't need additional checks there.

I've not written any new tests validating if this widget is shown based on the param, since the similar `isIterator` parameter also has no tests for this; happy to add some if you wish, but if so I'd appreciate a pointer to a similar existing test to help me understand this repo's testing approach.

---


echeung-amzn:
> since the similar `isIterator` parameter also has no tests for this

That's probably a mistake on my part; it'd be ideal to have some sort of basic snapshot test for this at least.

Something like [this test](https://github.com/cdklabs/cdk-monitoring-constructs/blob/f50190586d8fbf03fef66ec34f65603ec2aa0a70/test/monitoring/aws-kinesis/KinesisFirehoseMonitoring.test.ts#L21) that sets up the monitoring and ultimately snapshots the dashboards should be sufficient. Since that should already include the existing widgets, a new test would just set the new prop to `false` and the snapshot can confirm whether the widgets are indeed removed or not.

echeung-amzn (approved):
Thanks!

PR #629, opened 2025-03-28
fix(dynamo): adjusts DynamoTableGlobalSecondaryIndexMonitoring to respect localAlarmNamePrefixOverride

### LINKED ISSUE #628: DynamoTableGlobalSecondaryIndexMonitoring ignores localAlarmNamePrefixOverride
### Version

5.10.4

Given two dynamo tables, each with identically named GSIs, adding alarms (such as `addReadThrottledEventsCountAlarm`) will create a construct id name collision: 
```
Error: There is already a Construct with name '<globalPrefix>-<GSIName>-Throttled-Events-Warning' in AmazonMonitoringFacade [Monitoring]
```

This is not fixed by adding a localAlarmNamePrefixOverride.

If I get some time I'll put together a reproduction example.

### Expected behavior

I would expect the localAlarmNamePrefixOverride to allow updating the local part of the alarm name (GSI name).

### Actual behavior

Value is ignored.

### Other details

This occurs because [DynamoTableGlobalSecondaryIndexMonitoring](https://github.com/cdklabs/cdk-monitoring-constructs/blob/main/lib/monitoring/aws-dynamo/DynamoTableGlobalSecondaryIndexMonitoring.ts#L95) calls createAlarmFactory on it's parent scope, MonitoringFacade, instead of on it's own superclass, [Monitoring](https://github.com/cdklabs/cdk-monitoring-constructs/blob/main/lib/common/monitoring/Monitoring.ts#L67) which handles the alarm local name override.

haljarrett:
Fixes #628

This change passes the BaseMonitoringProps passed into DynamoTableGlobalSecondaryIndexMonitoring along to the super class constructor, which exposes `localAlarmNamePrefixOverride` to be consumed by Monitoring.createAlarmFactory. Additionally, this changes the construction of the alarm factory to use `createAlarmFactory` from the parent class (thus using `localAlarmNamePrefixOverride`, rather than that from the parent scope (MonitoringFacade, which doesn't have  `localAlarmNamePrefixOverride` available).

By doing so, we achieve the more expected behavior of having `localAlarmNamePrefixOverride` influence the local portion of the alarm name. By default, that portion is just the GSI name itself, and there are cases where this can cause collisions (e.g. two GSIs with the same name on different tables). Having  `localAlarmNamePrefixOverride` function as expected provides an escape hatch for this corner case without changing the default naming behavior. (i.e. including table name in the alarm name).

---

### DIFF
--- a/lib/monitoring/aws-dynamo/DynamoTableGlobalSecondaryIndexMonitoring.ts
+++ b/lib/monitoring/aws-dynamo/DynamoTableGlobalSecondaryIndexMonitoring.ts
@@ -59,7 +59,7 @@ export class DynamoTableGlobalSecondaryIndexMonitoring extends Monitoring {
     scope: MonitoringScope,
     props: DynamoTableGlobalSecondaryIndexMonitoringProps,
   ) {
-    super(scope);
+    super(scope, props);
 
     const namingStrategy = new MonitoringNamingStrategy({
       ...props,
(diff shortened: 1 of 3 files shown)


echeung-amzn (approved):
Thanks!

PR #632, opened 2025-04-01
feat: add ability to clone monitors

### LINKED ISSUE #631: Capability to create aggressive rollback alarms that mirror more conservative ticketing alarms
### Feature scope

All

Hello. I have a feature request for cdk-monitoring-constructs.

### User Story

I use CloudWatch alarms for two purposes:

1. Ticketing alarms file a high-severity issue in my ticket management software, triggering a page to the on-call engineer.
2. Rollback alarms signal that any in-progress or recently completed deployment should be reverted automatically.

As the owner of a complex application with many monitorable resources, I want to ensure regressions that result in an SLA-breach are automatically rolled back _before_ my ticketing action takes effect. For each monitor used in a high-severity ticketing action, I would like a companion monitor used for triggering pipeline rollbacks. The companion rollback monitor should use the same metrics and thresholds as the ticketing monitor, but with a more aggressive datapoints to alarm. For example, if I ticket when latencyP90 is greater than 1000ms for 15 minutes, then I should rollback when latencyP90 is greater than 1000ms for 5 minutes.

`MonitoringFacade` should introduce the capability to specify that all monitors belonging to a given disambiguator (e.g. "Critical") should be cloned to new monitors with a new different disambiguator (e.g. "Rollback") and mutated to a smaller number of datapoints to alarm. The interface for specifying this behavior should avoid requiring consumers to repeat each monitor threshold twice.

### Background Info

Tighter rules for rollback alarms can be a good strategy for healing from deployment regressions and reducing ticket counts.

For each of our applications, we set SLAs for high-severity issues. Generally speaking, if the threshold is breached for a short amount of time, we need not page our on-call. We only cut a ticket if the threshold is breached for several data points (e.g. 15 minutes). But if the threshold is breached shortly after a deployment, we needn't wait for so many data points: rollback immediately because the deployment likely caused the problem.

Rolling changes back sooner than our ticketing alarm also has the desirable benefit that we cut fewer tickets. If problematic deployments can be reverted automatically, then the system will heal itself without human intervention. There's no need to cut a high-severity ticket when the system is already healing. Therefore, if the wait duration for the ticketing alarm is longer than it takes for the rollback alarm to trigger and heal the application, then the ticketing alarm won't trigger and we avoid the unnecessary page. Only if the application fails to return to a healthy state would we need to page the on-call to investigate further.

### Possible Interface Idea

This is just an idea for how the changes to `MonitoringFacade` might look. It's just to spark discussion; I'm open to other suggestions.

We can build this feature in two parts:

#### Part One: Allow consumers to supply a generic "clone function"

The `MonitoringFacade` class can add a new `cloneAlarms()` method. When given a list of `AlarmWithAnnotation` objects and a TypeScript function, the `cloneAlarms()` method would apply the function on each alarm in the list. The developer can obtain an input list of alarms by calling the existing `createdAlarms()` or `createdAlarmsWithDisambiguator()` methods.

The clone function itself would accept an `AlarmWithAnnotation` instance and return a new `AddAlarmProps` instance that describes a new alarm to create. Developers can implement this function in any way they like by inspecting the input alarm, adjusting/tweaking values, and returning the specification for a new alarm.

Once the function generates a new list of `AddAlarmProps` objects, the new `cloneAlarms()` method would invoke the alarm factory to create those alarms and return them to the consumer.

```typescript
// Create a monitoring facade
const facade = new MonitoringFacade(stack, "MyAppFacade", { ... });
// Add a bunch of alarms
...
// Clone those alarms using a cloning-function
const criticalAlarms = facade.createdAlarmsWithDisambiguator("Critical");
const clones = facade.cloneAlarms(criticalAlarms, (a) => {
   // Define a new alarm that has values inspired by the original alarm
   // Adjust some of those values using arbitrary, user-provided logic
   return {
      ...a.alarmDefinition.addAlarmProps,
      actionsEnabled: false,
      disambiguator: "ClonedCritical",
      alarmDescription: "Cloned alarm of " + a.alarmDescription,
      // Bump the threshold a bit
      threshold: a.alarmDefinition.addAlarmProps.threshold * 1.1,
      // Tighten the number of datapoints a bit
      datapointsToAlarm: a.alarmDefinition.datapointsToAlarm - 1,
      // Keep the same number of evaluation periods
      evaluationPeriods: a.alarmDefinition.evaluationPeriods,
   }
});
```

#### Part Two: Provide a specialized-purpose "ScaleAlarm" function

To easily enable the use case describe above of creating aggressive rollback alarms by cloning more-conservative ticketing alarms, we can vend an implementation of the alarm cloning function. The function can be customized by consumers with scaling factors for `threshold`, `datapointsToAlarm`, and `evaluationPeriods`; scaling factors between 0.0 and 1.0 will result in more aggressive alarms. By providing this out-of-the-box, consumers can simply indicate how much more aggressive they want their alarms to be.

```typescript
// Clone those alarms using the scaling function
const criticalAlarms = facade.createdAlarmsWithDisambiguator("Critical");
const rollbackAlarms = facade.cloneAlarms(criticalAlarms, ScaleAlarms({
   disambiguator: "Rollback",
   thresholdMultiplier: 0.8,
   datapointsToAlarmMultiplier: 0.3,
   evaluationPeriodsMultiplier: 0.5,
}));
```

### Alternatives Considered

The following alternatives would technically work, but are not as desirable because they shift complexity to the consumer:

1. **Require application owners to specify thresholds twice for each monitor.** Technically speaking, consumers of MonitoringCDKConstructs can create aggressive rollback alarms already. They would just specify both alarms each time with a unique disambiguator. However, this approach isn't desirable to me. It requires additional code from the consumer, and there's no guarantee that the code is added consistently to every monitorable resource. In a large application, one could easily omit a rollback alarm or set a threshold inconsistently.
2. **Define alarms with the fewer datapoints to alarm and wrap them in a delay for ticketing.** Instead of specifying alarms using their ticketing data points, we could reverse the usage and expect consumers to specify the number of data points as the number required for rollback. These alarms would be directly used for deployment rollbacks and would have no ticketing action. We'd then wrap those alarms in a composite ticketing alarm and delay cutting the ticket until the source alarm has been active for a long period of time. This approach would work without having to specify thresholds twice, but it seems awkward for the developers using the feature. When setting application SLAs, our teams tend to think in terms of how long should it take to pull in on-call intervention. They would have to then divide that number by some factor and use that quotient in their code. This would be less user-friendly for our teams.

kfritz:
Closes #631

---

### Summary

The changes in this pull request implement [a feature to clone conservative ticketing alarms to be more aggressive for rollback alarms](https://github.com/cdklabs/cdk-monitoring-constructs/issues/631). `MonitoringFacade` introduces the capability to specify that all monitors belonging to a given disambiguator (e.g. "Critical") should be cloned to new monitors with a new different disambiguator (e.g. "Rollback") and mutated to a smaller number of datapoints to alarm.

There are two parts to this feature.

The `MonitoringFacade` class now has a `cloneAlarms()` method. When given a list of `AlarmWithAnnotation` objects and a TypeScript function, the `cloneAlarms()` method applies the function on each alarm in the list. The clone function itself takes an `AlarmWithAnnotation` instance and returns a new `AddAlarmProps` instance that describes a new alarm to create. Once the function generates a new list of `AddAlarmProps` objects, the `cloneAlarms()` method then invokes the alarm factory to create those alarms and return them to the consumer.

To easily enable the use case describe in #631 of creating aggressive rollback alarms by cloning more-conservative ticketing alarms, the PR includes an implementation of the alarm cloning function. The function can be customized by consumers with scaling factors for `threshold`, `datapointsToAlarm`, and `evaluationPeriods`; scaling factors between 0.0 and 1.0 will result in more aggressive alarms.

### Implementation Details

In order for the new `cloneAlarms()` method to create new alarms, it needs to obtain information about the original alarm that currently is not stored. Specifically, it needs all the inputs to `AlarmFactory`'s `addAlarm()` method: a `MetricWithAlarmSupport` and an `AddAlarmProps` instance. It also needs the original `AlamFactory` instance itself.

Currently, the `AlarmFactory` and its inputs are discarded after alarms are originally created. Therefore, this CR creates a place to hold onto those objects for later use. A new type called `AlarmCreateDefinition` stores the factory, the metric, and the source alarm props, and an instance of this object is added to `AlarmWithAnnotation`. Whenever an alarm is created in `AlarmFactory`, the creation definition is stored on the resulting alarm object. That way, an alarm clone function can access these original values.

We also create the `ScaleAlarms` clone function implementation. This code can perform the following scaling operations when cloning:
* **threshold scaling** - When a threshold is specified with a "greater than" comparison operator, the scaling factor is multiplied against the original threshold. For example, a scaling factor of 0.5 would half the source threshold value, creating a more aggressive threshold. For "less than" comparison operators, we subtract the scaling factor from 1 so that we make the lower-bound threshold more aggressive too.
* **datapointsToAlarm and evaluationPeriods scaling** - Scaling factors will be multiplied by the source alarm's datapoint values. When the scaling factor is less than 1, this causes the alarm to trigger sooner. In the case where the original alarm has a low number of datapoints such that scaling it down would be problematic, we attempt to reduce the period duration so that we can still alarm sooner.

### Testing

This PR includes new unit test cases for both a user-supplied custom clone function and the common case of using the alarm-scaling clone function. The unit test performs both fine-grained assertions and also a snapshot verification.

I also deployed a CloudFormation stack with this feature to a personal account and manually verified it created the expected alarms.

---

### DIFF
--- a/lib/common/alarm/ScaleAlarms.ts
+++ b/lib/common/alarm/ScaleAlarms.ts
@@ -1,0 +1,157 @@
+import { Duration } from "aws-cdk-lib";
+import { ComparisonOperator, Stats } from "aws-cdk-lib/aws-cloudwatch";
+import { AlarmCloneFunction, AlarmWithAnnotation } from "../..";
+
+const SupportedMetricPeriods = [
+  Duration.minutes(1),
+  Duration.minutes(5),
+  Duration.minutes(15),
+  Duration.hours(1),
+  Duration.hours(6),
+  Duration.days(1),
+  Duration.days(7),
+];
+
+/**
+ * Properties for configuring a function that clones alarms to be scaled by multiplication factors.
+ */
+export interface ScaleAlarmsProps {
+  /**
+   * The disambiguator to assign to the alarms cloned by the function.
+   */
+  readonly disambiguator: string;
+
+  /**
+   * A multiplication factor to apply to the threshold of the cloned alarms.
+   *
+   * Threshold scaling does not apply to thresholds on anomaly detection alarms.
+   *
+   * @default - 1.0
+   */
+  readonly thresholdMultiplier?: number;
+
+  /**
+   * A multiplication factor to apply to the datapointsToAlarm property of the cloned alarms.
+   *
+   * When configured with a value less that 1.0, the cloned alarm *may* use a metric with a more
+   * granular period so that the result will use a reasonable number of datapoints. In this case,
+   * the `datapointsToAlarm` and `evaluationPeriods` are increased by the same factor that the
+   * period is decreased. It will also scale `threshold` for metrics using the SUM or SAMPLE_COUNT
+   * stats.
+   *
+   * @default - 1.0
+   */
+  readonly datapointsToAlarmMultiplier?: number;
+
+  /**
+   * A multiplication factor to apply to the `evaluationPeriods` property of the cloned alarms.
+   * @default - Same as datapointsToAlarmMultiplier.
+   */
+  readonly evaluationPeriodsMultiplier?: number;
+}
+
+/**
+ * Creates a function that clones alarms to be more scaled using the multiplication factors
+ * provided. This function can be provided to the `cloneAlarms` method of
+ * {@link MonitoringFacade}.
+ *
+ * This can be used with scaling factors less than 1.0 to create more aggressive alarms, which
+ * can be useful to provide early detection and trigger change rollbacks.
+ *
+ * If the datapoints to alarm would be scaled to such a small number that it can't effectively be
+ * alarmed, the metric period will be reduced to the next smallest duration and the required
+ * datapoint count will scale up accordingly.
+ *
+ * @param props The properties for configuring the function.
+ * @returns A function that clones alarms and applies scaling factors.
+ */
+export function ScaleAlarms(props: ScaleAlarmsProps): AlarmCloneFunction {
+  return function (alarm: AlarmWithAnnotation) {
+    let thresholdMultiplierToUse: number = 1;
+    if (props.thresholdMultiplier) {
+      const comparisonOperator =
+        alarm.alarmDefinition.addAlarmProps.comparisonOperator;
+      switch (comparisonOperator) {
+        case ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD:
+        case ComparisonOperator.GREATER_THAN_THRESHOLD:
+          thresholdMultiplierToUse = props.thresholdMultiplier;
+          break;
+        case ComparisonOperator.LESS_THAN_OR_EQUAL_TO_THRESHOLD:
+        case ComparisonOperator.LESS_THAN_THRESHOLD:
+          thresholdMultiplierToUse = 1.0 / props.thresholdMultiplier;
+          break;
+        case ComparisonOperator.GREATER_THAN_UPPER_THRESHOLD:
+        case ComparisonOperator.LESS_THAN_LOWER_THRESHOLD:
+        case ComparisonOperator.LESS_THAN_LOWER_OR_GREATER_THAN_UPPER_THRESHOLD:
+          // Don't modify thresholds for anomaly detection models.
+          thresholdMultiplierToUse = 1;
+          break;
+        default:
+          throw new Error(
+            `Comparison operator ${comparisonOperator} is not supported by ScaleAlarms.`,
+          );
+      }
+    }
+    let cloneThreshold =
+      alarm.alarmDefinition.addAlarmProps.threshold * thresholdMultiplierToUse;
+
+    let cloneEvaluationPeriods =
+      alarm.alarmDefinition.evaluationPeriods *
+      (props.evaluationPeriodsMultiplier ??
+        props.datapointsToAlarmMultiplier ??
+        1);
+    let cloneDatapointsToAlarm =
+      alarm.alarmDefinition.datapointsToAlarm *
+      (props.datapointsToAlarmMultiplier ?? 1);
+
+    // If the tightened datapoints to alarm is now so small that we can't effectively alarm on it,
+    // reduce the metric period to the next smallest duration and scale up the datapoint count accordingly.
+    let clonePeriod = alarm.alarmDefinition.addAlarmProps.period;
+    if (cloneEvaluationPeriods < 2 || cloneDatapointsToAlarm < 2) {
+      const originalPeriod =
+        alarm.alarmDefinition.addAlarmProps.period ??
+        alarm.alarmDefinition.metric.period;
+
+      const nextSmallestPeriod = findNextSmallestPeriod(originalPeriod);
+      if (nextSmallestPeriod) {
+        clonePeriod = nextSmallestPeriod;
+        const datapointScaleFactor =
+          originalPeriod.toSeconds() / nextSmallestPeriod.toSeconds();
+        cloneEvaluationPeriods = cloneEvaluationPeriods * datapointScaleFactor;
+        cloneDatapointsToAlarm = cloneDatapointsToAlarm * datapointScaleFactor;
+
+        // If the metric uses SUM or N, we also need to scale the threshold down.
+        if ("statistic" in alarm.alarmDefinition.metric) {
+          if (
+            alarm.alarmDefinition.metric.statistic === Stats.SUM ||
+            alarm.alarmDefinition.metric.statistic === Stats.SAMPLE_COUNT
+          ) {
+            cloneThreshold = cloneThreshold / datapointScaleFactor;
+          }
+        }
+      }
+    }
+
+    cloneEvaluationPeriods = Math.max(1, Math.round(cloneEvaluationPeriods));
+    cloneDatapointsToAlarm = Math.max(1, Math.round(cloneDatapointsToAlarm));
+
+    return {
+      ...alarm.alarmDefinition.addAlarmProps,
+      disambiguator: props.disambiguator,
+      threshold: cloneThreshold,
+      evaluationPeriods: cloneEvaluationPeriods,
+      datapointsToAlarm: cloneDatapointsToAlarm,
+      period: clonePeriod,
+    };
+  };
+}
+
+function findNextSmallestPeriod(originalPeriod: Duration) {
+  for (let index = SupportedMetricPeriods.length - 1; index >= 0; index--) {
+    const current = SupportedMetricPeriods[index];
+    if (current.toSeconds() < originalPeriod.toSeconds()) {
+      return current;
+    }
+  }
+  return undefined;
+}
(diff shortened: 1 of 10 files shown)


lib/common/alarm/ScaleAlarms.ts
```diff
@@ -0,0 +1,143 @@
...
+      alarm.alarmDefinition.evaluationPeriods *
+      (props.evaluationPeriodsMultiplier ??
+        props.datapointsToAlarmMultiplier ??
+        1);
+    let cloneDatapointsToAlarm =
+      alarm.alarmDefinition.datapointsToAlarm *
+      (props.datapointsToAlarmMultiplier ?? 1);
+
+    // If the tightened datapoints to alarm is now so small that we can't effectively alarm on it,
+    // reduce the metric period to the next smallest duration and scale up the datapoint count accordingly.
```
echeung-amzn, line 94: The docs for `ScaleAlarms` should be more transparent about this sort of clamping behaviour, otherwise it may be surprising.

lib/common/alarm/ScaleAlarms.ts
```diff
@@ -0,0 +1,143 @@
...
+          thresholdMultiplierToUse = props.thresholdMultiplier;
+          break;
+        case ComparisonOperator.LESS_THAN_OR_EQUAL_TO_THRESHOLD:
+        case ComparisonOperator.LESS_THAN_THRESHOLD:
+          thresholdMultiplierToUse = 1.0 / props.thresholdMultiplier;
+          break;
+        case ComparisonOperator.GREATER_THAN_UPPER_THRESHOLD:
+        case ComparisonOperator.LESS_THAN_LOWER_THRESHOLD:
+        case ComparisonOperator.LESS_THAN_LOWER_OR_GREATER_THAN_UPPER_THRESHOLD:
+          // Don't modify thresholds for anomoly detection models.
```
echeung-amzn, line 72: The docs for `ScaleAlarms` should be more transparent about the expected behaviour with anomaly detection, otherwise it may be surprising.

Also:
```diff
- anomoly
+ anomaly
```

lib/facade/MonitoringFacade.ts
```diff
@@ -363,6 +376,36 @@ export class MonitoringFacade extends MonitoringScope {
...
+
+  /**
+   * Applies a cloning function to each of the given alarms, creating a new collection of alarms
+   * that are adjusted by the function.
+   *
+   * @param sourceAlarms The alarms that should be used as sources for the clones.
+   * @param cloneFunction A function that will accept a source alarm and determine whether and how a new alarm should be cloned from it.
+   * @returns The list of clone alarms.
+   */
+  cloneAlarms(
```
echeung-amzn, line 390: I'm wondering if it'd make sense for the methods like `createdAlarms` to later also include the newly created alarms via cloning?

I know those methods just get things from the `Monitoring` implementations right now, so we'd need to do something to handle that, but my initial impression is that it'd more intuitive at least (there's a similar gap with created composite alarms that's never been filled, but I'm not sure we should keep doing that).

test/facade/MonitoringFacade.test.ts
```diff
@@ -172,4 +173,230 @@ describe("test of defaults", () => {
...
+    const warningAlarms = facade.createdAlarmsWithDisambiguator("Warning");
+    expect(warningAlarms.length).toBe(expectedNumberOfWarningAlarms);
+
+    const expectedNumberOfClonedAlarms = 3;
+    expect(clones.length).toBe(expectedNumberOfClonedAlarms);
+    expect(clones.map((a) => a.action)).toEqual(
+      expect.arrayContaining(
+        Array(expectedNumberOfClonedAlarms).fill(noopAction()),
+      ),
+    );
```
echeung-amzn, line 280: No strong opinion here, but may a bit clearer to just write something like:

```typescript
clones.forEach(a => expect(a.action).toBe(noopAction());
```

test/facade/MonitoringFacade.test.ts
```diff
@@ -172,4 +173,230 @@ describe("test of defaults", () => {
...
+    expect(clones.map((a) => a.disambiguator)).toEqual(
+      expect.arrayContaining(
+        Array(expectedNumberOfClonedAlarms).fill("UpperBoundCritical"),
+      ),
+    );
+    clones.forEach((cloneAlarm) => {
+      const sourceAlarm = criticalAlarms.find(
+        (a) => a.alarmDefinition.metric === cloneAlarm.alarmDefinition.metric,
+      )!;
+      expect(cloneAlarm.alarmDefinition.addAlarmProps.threshold).toBeCloseTo(
```
echeung-amzn, line 290: Floating point arithmetic making it hard to be exact here?

test/common/alarm/ScaleAlarms.test.ts
```diff
@@ -0,0 +1,441 @@
...
+        Metrics: Match.anyValue(),
+        Threshold: Match.anyValue(),
+        TreatMissingData: Match.anyValue(),
+      },
+    }),
+  );
+  expect(Object.keys(templateRollbackAlarms).length).toBe(
+    expectedNumberOfCriticalAlarms,
+  );
+  // Expect cloned alarms to match the requested alarms in all ways
```
echeung-amzn, line 313: Nitpick: at some point I do feel like having too much logic in the tests can make tests riskier over having more static assertions.

echeung-amzn (commented):
Overall looks pretty solid, just left some relatively minor comments.

Could you also rebase while you're at it?

test/facade/MonitoringFacade.test.ts
kfritz: Yep, that's right.

test/common/alarm/ScaleAlarms.test.ts
kfritz: I like having some explicit expectations, as I think it's more clear than snapshot testing what the original author's behavior intent was. But there's admittedly a lot here; I can trim this down if you prefer.

I could drop lines 249-312 without losing sleep. I'd rather keep lines 313-395, which verify that the threshold, datapoints, and evaluation period scaling all took effect without disrupting the underlying metric. Does that seem reasonable?

lib/facade/MonitoringFacade.ts
kfritz: I agree that `createdAlarms` should include the clones.

For composite alarms, though, we have a type problem. The `createdAlarms` method returns an array of `AlarmWithAnnotation`, but the composites are all of type `CompositeAlarm`. How do you want to handle that?

echeung-amzn: I'm not really sure... but that's also tangential to this PR so we can leave that for now.

test/common/alarm/ScaleAlarms.test.ts
echeung-amzn: Yup, seems reasonable.

lib/common/alarm/ScaleAlarms.ts
echeung-amzn: More specifically -- can we include this in the main JSDocs for `ScaleAlarms` rather than just in a comment here?

echeung-amzn (approved):

Issue #391, opened 2023-07-04
Support for dashboard variables


echeung-amzn:
### Feature scope

All

### Describe your suggested feature

From https://aws.amazon.com/about-aws/whats-new/2023/06/amazon-cloudwatch-dashboard-variables/ :

---

We are excited to announce Amazon CloudWatch dashboard variables, a new experience that makes it easier for you to quickly navigate between different operational views by using variables as navigation to view different data sets for the same resource.

With dashboard variables, you can create drill-down filters that enable you to build multiple views within a single dashboard. You can create a single dashboard with multiple custom variables, where you can switch between data sets, and more efficiently reuse and manage a more efficient fleet of dashboards.

Custom labels become navigation on your dashboard that can be used to toggle between data sets in the same view. You can now define set of labels such as 'application environment', 'region' or a 'customer id' and switch between those data sets based on the selection of label by using dropdown selectors, radio buttons or input boxes.

The new dashboard variable experience is now available in all [AWS commercial regions](https://aws.amazon.com/about-aws/global-infrastructure/regions_az/) at no additional cost and you can start using it immediately. You can use a step by step guide in the CloudWatch custom dashboard user interface to configure and manage your variables, or directly from a CloudWatch custom dashboard JSON, by adding an array of "variables" to the dashboard definition.

To learn more about creating dashboard variables, please refer to our [documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_dashboard_variables.html).

---

https://github.com/aws/aws-cdk/commit/73f274193d4c687822d956742722ce5cecdf9173

PR #639, opened 2025-04-30
feat(dashboard): add support for dashboard variables

Laxenade:
Fixes https://github.com/cdklabs/cdk-monitoring-constructs/issues/391

Not sure if this change is really needed — I’ve been creating dashboards with variables by just grabbing the underlying dashboards and attaching variables that way. That said, it’s a nice cleanup and does make the code a bit cleaner. Totally fine to close the PR if you don’t think it’s worth it.

---


echeung-amzn:
Thanks, this feels like a good incremental improvement.

With this, the expectation is for folks that just use `MonitoringFacade` would need to pass their own `dashboardFactory`, which in turn can use this new `variables` prop set for it?

Laxenade:
> Thanks, this feels like a good incremental improvement.
> 
> With this, the expectation is for folks that just use `MonitoringFacade` would need to pass their own `dashboardFactory`, which in turn can use this new `variables` prop set for it?

Correct, unless we create another facade similar to the internal one that flattens the factory props into the facade prop

echeung-amzn (approved):

PR #641, opened 2025-05-01
feat: use dynamically determined latest Node.js runtime for Lambda handlers instead

echeung-amzn:
`NODEJS_LATEST` still points to Node.js 18, which is already EOL.

Related upstream issue: https://github.com/aws/aws-cdk/issues/28125

---


giantcow (approved):
Hopefully this will be the last time we have to do this kind of change... 🫠

PR #642, opened 2025-05-05
feat(alarm): add support for creating nested composite alarms

### LINKED ISSUE #237: Allow nested aggregate alarms.
### Feature scope

Alarm

Support for creating composite alarms containing other composite alarms. Some services may run into the CloudWatch payload and alarm limit for composite alarms. 

A workaround is to create the first layer with monitoringCdk and manually create the aggregate alarm using CloudWatch constructs. It would simplify this workflow if we could do something like call `createCompositeAlarmsUsingTag` multiple times.

### LINKED ISSUE #618: Enhancement: Support Custom Composite Alarm Expression in AlarmFactory.AddCompositeAlarmProps
### Feature scope

Cloudwatch

Here's a draft for the GitHub issue:

```markdown
## Title
Enhancement: Support Custom Composite Alarm Expression in AlarmFactory.AddCompositeAlarmProps

## Description
Currently, `AlarmFactory.AddCompositeAlarmProps` interface only supports a single `compositeOperator` (AND/OR) to be applied across all
 alarms in the composite alarm. This limitation prevents the creation of more complex alarm conditions that require mixed operators or NOT 
conditions.

### Current Behavior
The current implementation only allows expressions like:
- `(Alarm1 AND Alarm2 AND Alarm3)`
- `(Alarm1 OR Alarm2 OR Alarm3)`

### Desired Behavior
Support for custom composite alarm expressions would allow more sophisticated monitoring scenarios, such as:
- `(Alarm1 AND NOT (Alarm2 OR Alarm3 OR Alarm4))`
- `((Alarm1 AND Alarm2) OR (Alarm3 AND NOT Alarm4))`

### Use Case
Many monitoring scenarios require more complex logic than simple AND/OR operations. For example:
- Trigger an alarm when a service is down (Alarm1) but NOT during scheduled maintenance windows (Alarm2)
- Alert when system load is high (Alarm1) AND memory usage is critical (Alarm2) AND NOT during known batch processing periods (Alarm3)

### Proposed Solution
Add a new optional property to `AddCompositeAlarmProps`:
```typescript
interface AddCompositeAlarmProps {
  // existing properties
  compositeOperator?: CompositeAlarmOperator;
  // new property
  customExpression?: string;
}
```

When `customExpression` is provided, it would take precedence over `compositeOperator`, allowing users to define custom composite 
alarm rules using the CloudWatch composite alarm expression syntax.

### Additional Context
This enhancement would align better with CloudWatch's native composite alarm capabilities, which already support complex expressions 
through the AWS Console and CloudFormation.
```

This issue description outlines the current limitation, provides clear examples of the desired functionality, and suggests a 
backward-compatible solution. Feel free to modify it according to your specific needs or the project's issue template requirements.

### LINKED ISSUE #640: Being able to get/create composite alarms using tags
### Feature scope

Alarms

Currently, in https://github.com/cdklabs/cdk-monitoring-constructs/blob/main/lib/facade/MonitoringFacade.ts#L358-L374, the code calls `createdAlarmsWithTag()`, which intentionally excludes composite alarms. That behavior makes sense, and I understand there are valid cases where we might not want to include composite alarms. 

That said, I think it would be valuable to give users more flexibility and choices. Supporting nested composite alarms (i.e., a composite alarm that references other composite alarms)—is an important use case for advanced alarming. While there are workarounds, none are as clean or intuitive as having native support in the facade.

I also noticed that issues https://github.com/cdklabs/cdk-monitoring-constructs/issues/237 and https://github.com/cdklabs/cdk-monitoring-constructs/issues/618 are requesting similar functionality. So I wanted to ask, is there any blocker that prevents us from, say,  introducing an optional flag to include composite alarms in `createdAlarmsWithTag()`—with the understanding that the user assumes full responsibility for avoiding infinite recursion?

### PR DESCRIPTION
**Pull request by Laxenade**
Fixes #237
Fixes #618
Fixes #640

This commit adds support for creating nested composite alarms (i.e., including composite alarms within another composite alarm when using `createCompositeAlarmUsingTag`).

To avoid breaking the existing API or changing the current logic for how alarms are looked up during composite alarm creation, the feature is gated behind a flag, which is disabled by default.

---

### DIFF
--- a/API.md
+++ b/API.md
@@ -67735,9 +67735,11 @@ new LambdaFunctionEnhancedMetricFactory(metricFactory: MetricFactory, props: Lam
 | <code><a href="#cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricAvgInitDuration">enhancedMetricAvgInitDuration</a></code> | *No description.* |
 | <code><a href="#cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricAvgMemoryUtilization">enhancedMetricAvgMemoryUtilization</a></code> | *No description.* |
 | <code><a href="#cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricFunctionCost">enhancedMetricFunctionCost</a></code> | *No description.* |
+| <code><a href="#cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricInitDuration">enhancedMetricInitDuration</a></code> | *No description.* |
 | <code><a href="#cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricMaxCpuTotalTime">enhancedMetricMaxCpuTotalTime</a></code> | *No description.* |
 | <code><a href="#cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricMaxInitDuration">enhancedMetricMaxInitDuration</a></code> | *No description.* |
 | <code><a href="#cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricMaxMemoryUtilization">enhancedMetricMaxMemoryUtilization</a></code> | *No description.* |
+| <code><a href="#cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricMaxTotalMemory">enhancedMetricMaxTotalMemory</a></code> | *No description.* |
 | <code><a href="#cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricP90CpuTotalTime">enhancedMetricP90CpuTotalTime</a></code> | *No description.* |
 | <code><a href="#cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricP90InitDuration">enhancedMetricP90InitDuration</a></code> | *No description.* |
 | <code><a href="#cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricP90MemoryUtilization">enhancedMetricP90MemoryUtilization</a></code> | *No description.* |
@@ -67768,6 +67770,12 @@ public enhancedMetricAvgMemoryUtilization(): Metric | MathExpression
 public enhancedMetricFunctionCost(): Metric | MathExpression
 ```
 
+##### `enhancedMetricInitDuration` <a name="enhancedMetricInitDuration" id="cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricInitDuration"></a>
+
+```typescript
+public enhancedMetricInitDuration(): Metric | MathExpression
+```
+
 ##### `enhancedMetricMaxCpuTotalTime` <a name="enhancedMetricMaxCpuTotalTime" id="cdk-monitoring-constructs.LambdaFunctionEnhancedMetricFactory.enhancedMetricMaxCpuTotalTime"></a>
 
 ```typescript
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/common/alarm/AlarmFactory.ts [hunk 1]
```diff
@@ -821,9 +821,37 @@ export class AlarmFactory {
...
+
+  /**
+   * Creates a composite alarm from a collection of alarm base objects.
+   * This allows creation of composite alarms that include both metric alarms and other composite alarms.
+   *
+   * @param alarms Array of AlarmBase objects (can include both Alarm and CompositeAlarm)
+   * @param props Customization options for the composite alarm
+   * @returns Newly created composite alarm
+   */
+  addCompositeAlarmFromAlarmBases(
```
echeung-amzn, line 852: I wonder if the new method should also return an `AlarmWithAnnotation` like `addAlarm` does for better consistency?

Also not a fan of this method name, but nothing better immediately comes to mind. :\

lib/common/alarm/AlarmFactory.ts
```diff
@@ -821,9 +821,37 @@ export class AlarmFactory {
...
+   */
   addCompositeAlarm(
     alarms: AlarmWithAnnotation[],
     props: AddCompositeAlarmProps,
+    compositeAlarms: CompositeAlarm[] = [],
+  ): CompositeAlarm {
+    const alarmBases = [
+      ...alarms.map((alarm) => alarm.alarm),
+      ...compositeAlarms,
+    ] as AlarmBase[];
```
echeung-amzn, line 840: Is this cast necessary? I thought `CompositeAlarm` was already an `AlarmBase`.

lib/facade/MonitoringFacade.ts
```diff
@@ -178,7 +192,7 @@ export class MonitoringFacade extends MonitoringScope {
     | IDashboardSegment
     | IDynamicDashboardSegment
   )[];
-  protected readonly createdComposites: CompositeAlarm[];
+  protected readonly createdComposites: CompositeAlarmWithMetadata[];
```
echeung-amzn, line 195: Spitballing here, but I wonder if it'd be possible to have a dummy `Monitoring` segment that just acts as a composite alarm factory that the facade delegates to? 

Things like `createdAlarms()` would "just work" by returning everything, although things like `createdCompositeAlarms()` would need to filter first. (That said, I wonder if `createdAlarms()` _should_ return created composite alarms too? It may be more intuitive to do so.)

Not sure how weird it is to have a segment that doesn't actually do any widget things though.

lib/common/alarm/AlarmFactory.ts
Laxenade: Yeah, I'm not a fan of the names either. I wish I could just overload the method, but that's not an option here.

> I wonder if the new method should also return an `AlarmWithAnnotation`, like `addAlarm` does, for better consistency?

I'm not sure I follow. The existing method `addCompositeAlarm()` already returns a `CompositeAlarm`. Wouldn't we want the new method to align with that?

lib/common/alarm/AlarmFactory.ts
Laxenade: yeah, it's not needed, it was mostly just a visual cue. I can remove it.

lib/facade/MonitoringFacade.ts
Laxenade: I'll comment back on the dummy segment, I need to think about whether it makes sense.

To answer your question about whether `createdAlarms()` should return the created composite alarm: I don't think that's feasible if we want to maintain backward compatibility. Changing the return value of the method could break some downstream callers, especially if they're casting the return value for whatever reason. I'm inclined to avoid opening that can of worms for now.

Laxenade: Okay, I've tried a few different approaches ( `createdAlarm` should still return `AlarmWithAnnotation` to maintain backward compatibility, as I'm not convinced changing that is a good solution):

* I tried to move the new methods on creating composite alarms into `Monitoring` and then created a facade method to filter them. While this approach works, it feels a bit awkward. It seems to add complexity without significantly improving readability or anything really.
* I also tried creating a new subclass of `Monitoring` to house the new methods. This is more readable than the first option. However, as you pointed out, it results in an empty, purposely built segment that solely captures the composite alarm and doesn't render any widgets.

So, my thought is that creating a separate `Monitoring` is overkill in this case.

lib/common/alarm/AlarmFactory.ts
echeung-amzn: We can specify the type for the variable instead in that case (e.g. `const alarmBases: AlarmBase[] = [...];`.

Laxenade: Yeah, I removed it. Thinking about it, it's not really needed.

Laxenade:
Closed accidentally, reopen it

PR #661, opened 2025-08-14
feat(kinesisdataanalytics): Add alarms for Checkpoints

costmatt-amzn:
This PR adds checkpoint alarms for Kinesis Data Analytics monitoring.


lib/common/monitoring/alarms/KinesisDataAnalyticsAlarmFactory.ts
```diff
@@ -14,11 +20,15 @@ export interface FullRestartCountThreshold extends CustomAlarmThreshold {
   readonly maxFullRestartCount: number;
 }
 
+// Using standard ErrorCountThreshold and ErrorRateThreshold from ErrorAlarmFactory
```
echeung-amzn, line 23: Don't really need this comment?

lib/monitoring/aws-kinesisanalytics/KinesisDataAnalyticsMonitoring.ts
```diff
@@ -197,13 +233,17 @@ export class KinesisDataAnalyticsMonitoring extends Monitoring {
     });
   }
 
-  createNumberOfFailedCheckpointsWidget(width: number, height: number) {
+  createCheckpointFailuresWidget(width: number, height: number) {
```
echeung-amzn, line 236: This method name change is technically a breaking change since we intentionally keep them public so the class is reasonably easy to extend if desired.

If we really want the new name that makes more sense, we should at least retain another deprecated method with the old name that just ends up calling the new one.

echeung-amzn (changes requested):

lib/common/monitoring/alarms/KinesisDataAnalyticsAlarmFactory.ts
costmatt-amzn: leftover Claude and didn't notice, will remove

echeung-amzn (approved):

PR #663, opened 2025-08-18
fix:  monitorEc2Instances for instanceId list failing 

xendo:
Fixes #
In the previous PR where I added monitoring for network egress ingress I started setting color to 'Total Network In" unfortunately this is not caught during test or synthesis but only during deployment, and I only did e2e testing for ASG monitoring.

per-instance-network
per-instance
per-instance-network-multiple

network-monitoring-multiple

---


lib/monitoring/aws-ec2/EC2Monitoring.ts
```diff
@@ -133,11 +135,13 @@ export class EC2Monitoring extends Monitoring {
     for (const disambiguator in props.addNetworkOutTotalBytesExceedThresholdAlarm) {
       const alarmProps =
         props.addNetworkOutTotalBytesExceedThresholdAlarm[disambiguator];
-      const createdAlarms = this.networkOutSumMetrics.map((metric) => {
+      const createdAlarms = this.networkOutSumMetrics.map((metric, idx) => {
+        const indexedDisambiguator =
+          idx > 0 ? `${disambiguator}-${idx}` : disambiguator;
```
xendo, line 140: instead of index I tried passing instanceId from the MonitorFactoryStrategy, I was able to make it work with typescript, but hit this JSII issue: https://github.com/aws/jsii/discussions/3463

echeung-amzn (approved):

PR #686, opened 2025-12-09
feat(build)!: upgrade to JSII@5

### LINKED ISSUE #633: Upgrade jsii and TypeScript
### Feature scope

All

Hello friends.

Currently, cdk-monitoring-constructs is using an older version of jsii. Two years ago, [AWS released jsii-5.x](https://aws.github.io/jsii/compiler-and-rosetta-maintenance/) with the goal of enabling more TypeScript features. It also sets up a versioning scheme where the jsii version number would more-or-less follow the supported TypeScript version number.

As a developer seeking to implement features in cdk-monitoring-constructs, I would like to avail myself of these new language features. Additionally, jsii-1.x that the library currently uses is out-of-support by AWS, so it would be helpful to upgrade.

### Implementation considerations

For what it's worth, I took a look at what would be required to update cdk-monitoring-constructs to jsii-5.7.3 and TypeScript-5.7.3. _Most_ of the code works fine, except for [the `MonitoringAspect` features](https://github.com/cdklabs/cdk-monitoring-constructs/blob/main/lib/facade/MonitoringAspect.ts). These types fail to compile with jsii-5 because of the generic type declaration of [`MonitoringAspectType<T>`](https://github.com/cdklabs/cdk-monitoring-constructs/blob/main/lib/facade/IMonitoringAspect.ts). As stated in [the jsii documentation](https://aws.github.io/jsii/user-guides/lib-author/typescript-restrictions/#typescript-mapped-types):
> Parameterized types are not consistently supported in all supported target languages. Go support for generics is relatively limited (for good reasons) compared to TypeScript and most Object-Oriented languages, and the differences in generic semantics between TypeScript, C# and Java make it difficult to correctly represent such types in all those languages. As a consequence, `jsii` does not support declaring parameterized types.

Given the limitation described above, it makes sense why `MonitoringAspectType<T>` doesn't work with jsii-5.x. In fact, it shouldn't work in jsii-1.x either, and it's not clear to me why it is compiling without error in jsii-1.x. My best guess for what's happening in jsii-1.x comes from [this note in the jsii-5.x announcement](https://aws.github.io/jsii/compiler-and-rosetta-maintenance/#how-difficult-is-it-to-migrate-from-1x-to-50x):
> The 1.x release line silently ignores index signatures (these are hence unavailable in other languages), and the 5.0.x release starts explicitly rejecting these

That's not really talking about generic types, but I'm guess it's the same situation? Perhaps jsii-1.x is just dropping the monitoring aspect feature in non-TS/JS languages, while jsii-5.x explicitly rejects it. If so, we'd need to decide how to handle this situation if we upgrade to jsii-5.x:
* Find a way to keep aspect monitoring for TS/JS-only. I tried putting `@jsii ignore` on all the types, but it didn't work the way I expected -- could use some help from someone more familiar with jsii.
* Delete the aspect monitoring feature completely. Clearly, that's a regrettable option, but could be worth discussing.

arkon:
This removes some APIs/types that had replacements added previously in order to avoid using generics and tuples, which JSII@5 does not support.
    
The BaseMetricFactory might affect other extensions of that abstract class outside of this library, which would need to be confirmed by reviewers.

This does not address Rosetta doc errors, which are still disabled.

This also just uses the default JSII version that Projen currently specifies, which despite being documented as `~5.8.0` gives `~5.7`.

Given this is technically breaking changes to some APIs, this should include a major version update if merged (and grouped with other breaking changes, if any).

Fixes #633

---

Related docs:
- https://aws.github.io/jsii/compiler-and-rosetta-maintenance/

---


lib/facade/MonitoringAspect.ts
```diff
@@ -84,8 +84,8 @@ export class MonitoringAspect implements IAspect {
     }
   }
 
-  private getMonitoringDetails<T>(
-    aspectOptions?: MonitoringAspectType<T>,
+  private getMonitoringDetails<T extends Record<string, any>>(
+    aspectOptions?: BaseMonitoringAspectType & { props?: T },
```
giantcow: You mentioned removing generics, was that specifically for generics in interfaces or do we need to need to remove *all* generics (ie the `T` type you have on the function)

API.md
giantcow: Not looking at this in depth, loading the diff almost crashed my firefox tab 😆 Since this is auto generated I'm not too pressed

giantcow (commented):
Not sure why the auto-queue task is failing... there's not a lot of context in the logs

https://github.com/cdklabs/cdk-monitoring-constructs/pull/686/checks#step:2:131

API.md
arkon: tbh I never review it anyway.

lib/facade/MonitoringAspect.ts
arkon: https://aws.github.io/jsii/user-guides/lib-author/typescript-restrictions/#typescript-mapped-types

It's specifically usage in what ends up being part of the constructs' APIs. Internal usage is still fine (as evidenced by the satisfied build without JSII complaining about it), since that's still just always run within a Node.js context even if the consumer isn't directly using it.

README.md
```diff
@@ -549,6 +549,53 @@ const rollbackAlarms = monitoring.cloneAlarms(criticalAlarms, ScaleAlarms({
 }));
 ```
 
+## Upgrading from v9
```
arkon: I'd probably just put this in the actual GitHub release's body, but up to you.

giantcow (approved):

PR #687, opened 2025-12-10
feat: introduce replacement types for aspect

arkon:
This is extracted from #686 and an incremental step towards upgrading to JSII@5.x.

This introduces new types to replace the usage of the generic type that needs to be removed since JSII@5.x doesn't support generics as part of constructs' APIs.

This does kind of spam new interfaces into the API, but I think that's an acceptable trade-off (the alternative is just removing this aspect, or moving it to a TS-only library).

`MonitoringAspectType` would be removed as part of a later breaking change.

---

### DIFF
--- a/lib/facade/IMonitoringAspect.ts
+++ b/lib/facade/IMonitoringAspect.ts
@@ -33,7 +33,7 @@ import {
 /**
  * @deprecated Extend {@link BaseMonitoringAspectType} instead.
  *
- * This type is not compatable with JSII@5.x.
+ * This type is not compatable with JSII@5.x and will be removed in the next major version
  */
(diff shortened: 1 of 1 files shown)


lib/facade/IMonitoringAspect.ts
```diff
@@ -30,52 +30,173 @@ import {
   WafV2MonitoringOptions,
 } from "../monitoring";
 
-export interface MonitoringAspectType<T> {
+/**
+ * @deprecated Extend {@link BaseMonitoringAspectType} instead.
+ *
+ * This type is not compatable with JSII@5.x.
```
giantcow:
```suggestion
 * This type is not compatable with JSII@5.x and will be removed in the next major version
```

giantcow (approved):

PR #691, opened 2025-12-18
feat(alarm): Support AT_LEAST when creating composite alarm

Laxenade:
Adding support for creating composite alarms using the new `AT_LEAST` operator. The corresponding change in cdk-lib has not been merged https://github.com/aws/aws-cdk/pull/36100, and historically, this package lags behind cdk-lib by several months. Therefore, instead of using the new construct in cdk-lib, we will compose the alarm string ourselves. This is not a big deal since most of the code in the PR is still necessary. The only aspect that may change once we have the construct in cdk-lib is likely the last few lines in `determineCompositeAlarmRule`.

---


arkon (commented):
LGTM, although I don't have write permissions.

giantcow (changes requested):

bigbadtrumpet (approved):
Changes look good to me! Appreciate the contribution

giantcow:
@Laxenade my apologies on missing this

Looks like the docs need to be updated, the build is failing https://github.com/cdklabs/cdk-monitoring-constructs/actions/runs/22203969364/job/64223559747?pr=691

PR #738, opened 2026-04-29
feat!: Upgrade CDK

giantcow:
Upgrades to CDK 2.251.0 and constructs 10.5.0

Fixes #
- https://github.com/cdklabs/cdk-monitoring-constructs/security/dependabot/18
- https://github.com/cdklabs/cdk-monitoring-constructs/security/dependabot/14
- https://github.com/cdklabs/cdk-monitoring-constructs/security/dependabot/20

---


.projenrc.ts
```diff
@@ -1,7 +1,7 @@
 import * as cdklabs from "cdklabs-projen-project-types";
 import { javascript, github, DependencyType, ReleasableCommits } from "projen";
 
-const CDK_VERSION = "2.160.0";
+const CDK_VERSION = "2.251.0";
```
arkon, line 4: Increases to the minimum peer dependency versions were traditionally considered breaking changes and also needed to be aligned internally, FYI.

giantcow: That's fair- On one hand I'd like to do this major upgrade since there has been a lot of changes since- but I would also find it annoying to have back-to-back major releases :thinking: 

I'll think on it some more before merging

PR #759, opened 2026-07-25
chore!: Upgrade to CDK v2.260.0

giantcow:
Been putting this off for too long, the list of security vulns only grows...

Had to adjust anomolgy detection alarms slightly with new aws-cdk-lib requirements. Other then that it's just pass-through updates to snapshots

---


arkon:
Could consider [2.261.0](https://github.com/aws/aws-cdk/releases/tag/v2.261.0) to potentially make leveraging metrics facades easier in the future.

giantcow:
> Could consider [2.261.0](https://github.com/aws/aws-cdk/releases/tag/v2.261.0) to potentially make leveraging metrics facades easier in the future.

ehhh, it's [still in alpha](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk/metrics-facade-alpha) so we'll need a major version bump anyways once it goes GA

test/monitoring/aws-secretsmanager/SecretsManagerSecretMonitoring.test.ts
```diff
@@ -107,8 +107,11 @@ test("snapshot test", () => {
 });
 
 test("each stack in an app gets its own publisher instance", () => {
-  const app = new App();
-  for (const stack of [new Stack(app, "Stack1"), new Stack(app, "Stack2")]) {
+  const templates: Template[] = [];
+  for (const stackName of ["Stack1", "Stack2"]) {
```
miloszwatroba, line 111: this test is no longer representative of its description - "each stack in an app", while after the change we create new app for every stack.

package.json
```diff
@@ -72,8 +72,8 @@
     "typescript": "^4.9.5"
```
miloszwatroba, line 72: Q: shall we also bump TS? constructs and aws-cdk-lib build `~5.9.3` anyway

lib/common/monitoring/alarms/AnomalyDetectingAlarmFactory.ts
```diff
@@ -26,12 +29,21 @@ export class AnomalyDetectingAlarmFactory {
     disambiguator: string,
     props: AnomalyDetectionThreshold,
   ) {
-    return this.alarmFactory.addAlarm(metric, {
+    const alarmMetric =
+      metric instanceof AnomalyDetectionMathExpression
```
miloszwatroba, line 33: nit:  we have similar logic in other places, e.g. https://github.com/giantcow/cdk-monitoring-constructs/blob/main/lib/common/metric/MetricFactory.ts#L203 or even aws-cdk's [Metric#anomalyDetectionFor](https://github.com/aws/aws-cdk/blob/3c7254afd687db8ba70ea29a9730bad838adcac9/packages/aws-cdk-lib/aws-cloudwatch/lib/metric.ts#L411)

package.json
giantcow: I don't see why not, I'll throw it in

PR #770, opened 2026-09-05
feat(alarm): use single-glyph Unicode comparison operators in default alarm annotations

ketsugi:
## What

`DefaultAlarmAnnotationStrategy` passes `Alarm.toAnnotation()` through unchanged, so alarm annotation labels render comparison operators as the two-character ASCII sequences `>=` and `<=` (e.g. `TPS <= 0 for 1 datapoints within 5 minutes`). This change replaces them with the single Unicode glyphs `≥` and `≤` in the rendered label. Bare `>` and `<` are unchanged.

The substitution only matches space-padded operators, so metric names containing these character sequences are not affected.

## Why

The single glyphs are more readable on dashboard widgets, where annotation labels are often truncated and visually dense.

## Testing

- New unit test `test/common/alarm/IAlarmAnnotationStrategy.test.ts` covering all four comparison operators and asserting the ASCII digraphs no longer appear.
- Full test suite passes (82 suites, 416 tests); 3 snapshots updated in `LambdaFunctionMonitoring.test.ts.snap` where the diff is exactly `TPS <= 0` → `TPS ≤ 0`.

## Notes for reviewers

This changes rendered dashboard output for consumers using the default annotation strategy. Happy to gate it behind an opt-in prop on the strategy if you'd prefer the current rendering stay the default.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license.

### DIFF
--- a/lib/common/alarm/IAlarmAnnotationStrategy.ts
+++ b/lib/common/alarm/IAlarmAnnotationStrategy.ts
@@ -83,12 +83,33 @@ export abstract class FillingAlarmAnnotationStrategy
 }
 
 /**
- * Default annotation strategy that returns the built-in alarm annotation.
+ * Mapping of two-character comparison operator sequences (as rendered by
+ * Alarm.toAnnotation) to their single-glyph Unicode equivalents.
+ */
+const OPERATOR_GLYPHS: Record<string, string> = {
+  ">=": "≥",
+  "<=": "≤",
+};
+
+/**
+ * Default annotation strategy that returns the built-in alarm annotation,
+ * with two-character comparison operators (>=, <=) replaced by their
+ * single-glyph Unicode equivalents (≥, ≤) for readability.
  */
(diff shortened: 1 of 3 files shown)


arkon:
May want to check if this results in any annoyances when diffing against deployed templates. IIRC the library used to do something like this (or for other glyphs), but ended up moving away from it because running `cdk diff` would report a diff every time.

cf. https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1220

ketsugi:
`≥` and `≤` are both ASCII glyphs, though

arkon:
I think I'm either thinking of [this old PR](https://github.com/cdklabs/cdk-monitoring-constructs/pull/60) or something that was internal prior to that.

Either way if it doesn't show up incorrectly mangled in diffs (I don't actually know what CloudFormation handles, it may not actually include extended ASCII), then no worries; just a vague memory when I saw this PR. 👍

